PR #1042, opened 2014-05-22
Trigger an error when a required include fails to be included...

al-the-x:
I encountered this while attmepting to deploy a Roots-based theme to @WPEngine via git. A global `.gitignore` rule omitted the contents of `roots/lib/` from the index, so none of the `require_once` lines from `functions.php` worked. The `locate_template()` function can `require` the file, given appropriate flags, and returns an empty string if it can't find it. More useful would be an error message when this occurs. Whether this deserves `E_USER_ERROR`, `E_USER_WARNING` or `E_USER_NOTICE` is a subject for debate. Please advise.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -2,18 +2,29 @@
 /**
  * Roots includes
  */
-require_once locate_template('/lib/utils.php');           // Utility functions
-require_once locate_template('/lib/init.php');            // Initial theme setup and constants
-require_once locate_template('/lib/wrapper.php');         // Theme wrapper class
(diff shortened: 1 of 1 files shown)


kuldeepphp:
Hello Every one i got the same error 

My php version is 5.3.5.
Please advice me.

chrisatomix:
@kuldeepphp This is not a support forum, you should ask for assistance on the Roots Discourse:
https://discourse.roots.io

Issue #1192, opened 2014-10-30
Move lib/*.php functions into classes


alexsomeoddpilot:
I've noticed there's a lot of global functions in the `lib/*.php` files which deal with similar sets of data. Is there a reason they haven't been wrapped up into classes to deal with the data more cleanly and systematically? Is it for compatibility with PHP <5.3?

PR #1248, opened 2015-01-10
Add namespace

retlehs:


austinpray:
o͡͡͡͡͡͡͡͡͡͡͡͡͡͡╮(｡❛ᴗ❛｡)╭o͡͡͡͡͡͡͡͡͡͡͡͡͡͡

retlehs:
ref #1192

swalkinshaw:
We should use `__NAMESPACE__` for every namespace reference within the file so avoid repeating them.

retlehs:
done - [L68 nav.php](https://github.com/roots/roots/blob/90979457de1170a4b12c1ed8f5df3bb3ea021e71/lib/nav.php#L68) is fine, right? and nothing needs to be changed in any template files?

swalkinshaw:
[L68 nav.php](https://github.com/roots/roots/blob/90979457de1170a4b12c1ed8f5df3bb3ea021e71/lib/nav.php#L68) is correct yeah since it's referencing another namespace.

PR #1255, opened 2015-01-15
Use new manifest format

austinpray:
Simplifies and consolidates gulp tasks

Adds a --production flag to disable source maps


austinpray:
I temporarily removed CSSWring. We need https://github.com/hail2u/node-csswring/issues/47 to get resolved. I'm trying to help get that out the door.

austinpray:
Updated to use pleeease. Please is a wrapper for postcss that allows us to ignore stuff like this.

PR #1256, opened 2015-01-15
Update README.md

gegere:
Sudo might be needed needed with most systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Sage uses [gulp](http://gulpjs.com/) as its build system and [Bower](http://bowe
 
 From the command line:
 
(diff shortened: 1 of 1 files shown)


austinpray:
You do not want to be using sudo with npm commands. That is a gateway to a world of heartache and ruin. 

The most ideal situation is one uses NVM. If that is not possible then you need to make sure your file permissions are correct: https://stackoverflow.com/questions/16151018/npm-throws-error-without-sudo

gegere:
This is a very good tip and something I have never had explained. Would think a note should be added to the repo to help other who might look at `sudo` as a solution?

As I review files located in `~.npm` a update on ownership is definitely needed.

austinpray:
@gegere I am definitely a fan of briefly mentioning nvm. For people that are very new to node tools this could save them a lot of trouble. 
cc: @retlehs

PR #1258, opened 2015-01-15
Move .content markup out of PHP

leopuleo:
Move forward an agnostic framework:removed Bootstrap classes out from
PHP.

### DIFF
--- a/assets/src/less/layouts/_general.less
+++ b/assets/src/less/layouts/_general.less
@@ -1,5 +1,9 @@
 // Content wrapper
-.wrap { }
+.wrap { 
+  .content {
+    .make-row();
+  }
+}
 
 // Main content area
 .main {
--- a/base.php
+++ b/base.php
@@ -13,7 +13,7 @@
   ?>
 
   <div class="wrap container" role="document">
-    <div class="content row">
+    <div class="content">
       <main class="main" role="main">
         <?php include roots_template_path(); ?>
       </main><!-- /.main -->


JulienMelissas:
Looks good to me :+1:

retlehs:
your gulp branch is outdated (please use the latest 8.0.0 branch) so this PR can't be automatically merged

very similar to #1158, which was closed for now. we might want to hold off for the same reason

i also prefer @CFXd's slightly different version:

```less
.wrap {
  > .content {
    .make-row();
  }
}
```

leopuleo:
Thanks @retlehs, I will merge the latest 8.0.0 into my gulp branch.
I agree with @CFXd solution.

Question: Is it time to remove all Bootstrap classes?

austinpray:
I do hate to make things complicated with another opinion but:

```less
.wrap {
  > .content {
    background: blue;
  }
}
``` 

compiles to 

```css
.wrap > .content {
  background: blue;
}
```

I do not recommend this because of the high specificity of selectors with `>`. Changing the background of that .content element is now complicated. It should really be as simple as:

```css
.content {
  background: blue;
}
```

That way it is easy to override and theme out.

austinpray:
@leopuleo And I would say yes to all help weaning roots off of bootstrap. Success looks like: I should be able to use the same template files with both bootstrap and foundation.

leopuleo:
@austinpray :+1:

leopuleo:
Having a closer look to this issue, moving towards an agnostic theme, so users can easily choose between different frameworks (Bootstrap, Foundation...but also Semantic UI - why not) I figured out how difficult is, because:

* Different framework means different HTML structure, especially regarding menus or forms etc..
* Frameworks use different classes. A Bootstrap user can assert that Bootstrap classes are more semantic than Foundation ones and the opposite...still, both are :smile:. 
The best would be to choose third classes to extend with less or sass.
* Frameworks use different attributes: for example Boostrap navbar toogle has `data-toggle=` attribute, Foundation not...

I'm discovering something obvious, I know, but what do you guys think?
8.0.0 is on the way and you are doing an incredible job with it, but what are your wills for the future?

Cheers guys!

JulienMelissas:
@leopuleo - it's definitely going to be a challenge. I think the best thing is that the Yeoman generator will be able to include or not include files conditionally depending on what framework you might want to use (and can help walk you through it). I agree there are places in the markup where things might not be compatible, and while we can do our best to move things into Less/Sass, there will always be things that we simply can't.

austinpray:
@leopuleo it is not going to be the easiest thing in the world, but by no means is it insurmountable.

>- Different framework means different HTML structure, especially regarding menus or forms etc..

>- Frameworks use different attributes: for example Boostrap navbar toogle has data-toggle= attribute, Foundation not...

I am not concerned with the fact that foundation/bootstrap/whatever have different attributes for things like forms, sliders, tabs, modals. Roots includes no forms, sliders, tabs, modals by default, so how those are coded is completely up to the user. This would probably be the header menu and that is it. With the generator coming out this will not be hard to manage at all.

> - Frameworks use different classes. A Bootstrap user can assert that Bootstrap classes are more semantic than Foundation ones and the opposite...still, both are :smile:. The best would be to choose third classes to extend with less or sass.

We are on the same page here. We should not be putting framework-specific classes in the default templates.

Throwback to: http://www.csszengarden.com/

leopuleo:
Thanks guys for your reply!
As I see, removing all the Bootstrap classes from php to less/sass files is the next step.
I'm starting to work on it in my spare time...for now, thank you!

austinpray:
Thank you @leopuleo !

PR #1261, opened 2015-01-16
Adding live reload triggers to JS and CSS changes.

JulienMelissas:
Adding live reload triggers to JS and CSS changes.


austinpray:
`$.livereload` goes after `dest`. Otherwise you are creating a race condition.

JulienMelissas:
^ that's dumb that it failed, not my fault

PR #1267, opened 2015-01-18
[8.0.0] Use short echo syntax, bump required PHP version to 5.4

retlehs:
See http://php.net/manual/en/function.echo.php


swalkinshaw:
Requires a `README` update for the setting or min PHP requirement.

retlehs:
updated

PR #1271, opened 2015-01-20
Use the theme styles for the editor stylesheet

retlehs:
i think this will work fine. couple basic examples from working on the new roots.io theme:

image

image


austinpray:
This is so pro

JulienMelissas:
Been doing this for a while. Works well 90% of the time, great call and +1
for the import so you can override styles.

On Tuesday, January 20, 2015, Austin Pray <notifications@github.com> wrote:

> This is so pro
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/roots/roots/pull/1271#issuecomment-70658005>.
>


-- 
- Julien

austinpray:
@JulienMelissas just curious so I can make the judgement call in the future: when does it fail?

JulienMelissas:
There are some websites that have dark backgrounds that I've had to customize the stylesheet a good bit. Also some weird font stuff or custom styling if you rely on shortcodes for that (because they don't render out for the backend)... That's about it!

kjprince:
Awesome.

Issue #1188, opened 2014-10-28
Google analytics showing up in development environment


shockalotti:
In the scripts.php file I see this block of code:

```
if (GOOGLE_ANALYTICS_ID && (WP_ENV !== 'production' || !current_user_can('manage_options'))) {
  add_action('wp_footer', 'roots_google_analytics', 20);
}
```

If I'm not mistaken I think it should be:

```
if (GOOGLE_ANALYTICS_ID && (WP_ENV == 'production' || !current_user_can('manage_options'))) {
  add_action('wp_footer', 'roots_google_analytics', 20);
}
```

PR #1272, opened 2015-01-21
Improve Google Analytics conditional loading

swalkinshaw:
The current logic for condtionally including the GA snippet is confusing (see #1232, #1188 and [here](http://discourse.roots.io/t/google-analytics-condition-seems-off/2778/4)) and bit inconsistent. In development you would see logging when the snippet wasn't included but this wouldn't happen if you were a logged in admin.

Now there's only two cases:

1. Production non admin user: GA snippet is loaded.
2. Anything else (development or admin user): log output

### DIFF
--- a/lib/assets.php
+++ b/lib/assets.php
@@ -84,23 +84,25 @@ add_action('wp_head', __NAMESPACE__ . '\\jquery_local_fallback');
  *
  * Cookie domain is 'auto' configured. See: http://goo.gl/VUCHKM
  */
-function google_analytics() { ?>
-<script>
-  <?php if (WP_ENV === 'production') : ?>
-    (function(b,o,i,l,e,r){b.GoogleAnalyticsObject=l;b[l]||(b[l]=
-    function(){(b[l].q=b[l].q||[]).push(arguments)});b[l].l=+new Date;
-    e=o.createElement(i);r=o.getElementsByTagName(i)[0];
-    e.src='//www.google-analytics.com/analytics.js';
-    r.parentNode.insertBefore(e,r)}(window,document,'script','ga'));
-  <?php else : ?>
-    function ga() {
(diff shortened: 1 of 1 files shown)


austinpray:
FYI GA goes right before the end of the head tag. Analytics.js is non-blocking.

JulienMelissas:
👍

QWp6t:
Yeah, I'm down with that. The only thing worth noting is what I said in #1232:

> The problem with this, however, is that you're now loading that helper function in production for all admins, something that it wasn't intended to be for.

In my view, that logger is something that should only be exposed to developers. The reason I have it only showing to admins in non-production is because typically only clients and the developer would be viewing the site in non-production. Clients would be non-admins, and developers would be admins.

Like I said, I'm down with the change, though. It's just not the initial reason I added the logger.

austinpray:
BTW: A naked console.log like that will break all versions of Internet explorer below 10 and any old browsers that don't expose a console property on window. This will cause trouble for people trying to browser test on dev and staging and such.

swalkinshaw:
@QWp6t how many non-developers will be looking at the console though? I agree with your reasoning but I think that would be so uncommon that it's fine as is.

rtpHarry:
Bit busy to sort a pull request at the moment but noticed the GA snippet that it suggests you use has changed slightly now:

    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
    })(window,document,'script','//www.google-analytics.com/analytics.js','ga');

PR #1273, opened 2015-01-21
Begin to move theme vars into a central location

emaildano:
Proposal to move theme vars into a central location. This idea reflects most frameworks and would separate functions which aren't often edited and their values which are.


swalkinshaw:
Don't think this makes sense right now. There's only 2 "vars" and doesn't it make logical sense for them to be defined in a `config.php` anyway?

PR #1289, opened 2015-01-28
browserSync

guix77:
Like said in https://github.com/roots/roots/issues/128, here's a first try with browserSync.

In gulpfile.js people might need to change the proxy from *localhost* if they are using another local host name.

I tested with *.php changes and it works, but I couldn't test with styles and scripts (I'm getting independent errors).

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -24,5 +24,8 @@
     "modernizr.js": {
       "bower": ["modernizr"]
     }
+  },
+  "config": {
+    "devUrl": "example.dev"
   }
 }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -131,7 +131,7 @@ var writeToManifest = function(directory) {
 // Start the server
 gulp.task('browser-sync', function() {
     browserSync({
-        proxy: "aubonsite"
+        proxy: "localhost"
     });
 });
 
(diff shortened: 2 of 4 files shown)


austinpray:
Almost! I fixed the issues here: https://github.com/guix77/roots/pull/1

Merge that pull and it should be good to go.

emaildano:
@austinpray @guix77:

Using browserSync start on watch but I'm having one issue using it. When I run `gulp watch` my browser opens a new tab with `localhost:3000` even though I specified the config devURL in `manifest.json`

Any suggestions?

austinpray:
@emaildano localhost:3000 is the browsersync proxy address. Browsersync is proxying your dev server, sprinkling sugar ontop of it, and then serving it to you at localhost:3000

emaildano:
@austinpray gotcha. I'm new to browserSync. Is there anyone to not have it mask the domain name? I like my URLs with just a light sprinkle of sugar.

I'm looking through their docs now as-well.

guix77:
Guys, don't forget to check browserSync external URL in the log when you launch gulp watch, like

    [BS] External URL: http://192.168.0.26:3000

Fire up that URL on your smartphone, tablet, desktop, smartTV, and then watch the magic ! Scroll, navigate, inject CSS, ... :)

austinpray:
:sunglasses:

PR #1291, opened 2015-01-28
remove lodash from gulpfile and package.json

Tidal-Wave:
Speaking with @austinpray, lodash is not needed anymore.


austinpray:
Yep! And next: https://github.com/austinpray/asset-builder/issues/13

PR #1294, opened 2015-01-29
Fix node-sass and npm cache on travis

austinpray:
`cache: npm` isn't even a valid config anyway

https://stackoverflow.com/questions/15393821/npm-err-cb-never-called/15483897#15483897
https://github.com/sass/node-sass/issues/638

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,18 @@
 sudo: false
 language: php
-cache: npm
+cache:
+  directories:
+  - node_modules
 php:
 - '5.6'
 - '5.5'
 - '5.4'
 
 before_install:
+  - npm cache clean
+  - npm prune
   - npm install -g bower gulp jscs
-  - npm install
+  - npm update
   - pyrus install pear/PHP_CodeSniffer
   - phpenv rehash
 


austinpray:
screenshot 2015-01-29 01 25 04

swalkinshaw:
Cache `node_modules`: http://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/

montchr:
Looks like that's only available for private repos currently (second to last sentence). :(

montchr:
At least, according to that page specifically.

swalkinshaw:
It's confusing but it should work on their new container based workers. See http://docs.travis-ci.com/user/workers/container-based-infrastructure/:

> Jobs running on container-based infrastructure:
>
1. start up faster
2. allow the use of caches for public repositories
3. disallow the use of sudo, setuid and setgid executables

montchr:
Wow this is really confusing… because I think travis-ci.com is supposed to be for private repos only… it says so in the sidebar, but they really don't do a good job of indicating which version of Travis CI you're looking at… nothing in the header, just the sidebar.

ntwb:
Yes, if your using container based infrastructure you can now use caching.

Once you have it configured you can view the cache at e.g. https://travis-ci.org/roots/roots/caches (from memory you'll only see this if signed in, this isn't public)

I'm quite out of date on Roots, let alone 8.0 Sage but in the case of Roots/Sage I think you'll only be able to cache the `/node_modules` directory, see http://docs.travis-ci.com/user/caching/#Arbitrary-directories

For even more title vs content ambiguity:
> http://docs.travis-ci.com/user/caching/#Things-not-to-cache
> The cache's purpose is to make installing language-specific dependencies easy and fast, so everything related to tools like Bundler, pip, Composer, npm, Gradle, Maven, is what should go into the cache.

austinpray:
The correct syntax to cache the node modules is this:
```yml
cache:
  directories:
  - node_modules
```

austinpray:
screenshot 2015-01-29 11 00 28

ntwb:
If your `node_modules` folder was actually cached then `npm install` wouldn't actually install the npm modules again would it? See https://travis-ci.org/austinpray/roots/jobs/48786078#L198

I think the directory you want cached is `/home/travis/build/austinpray/roots/node_modules`
See https://travis-ci.org/austinpray/roots/jobs/48786078#L201

austinpray:
> If your node_modules folder was actually cached then npm install wouldn't actually install the npm modules again would it?

Below is the thought process as to why I am using npm install right now. I'm trying to work out whether we should keep using npm install or switch to npm update.

## Discussion

### How to build a node app

For anyone not familiar with how dependency management on production node apps work:

I generally base the build process for my apps based on the heroku node buildpack.

1. A regular npm install (first build; default scenario)
2. Copy existing node_modules from cache, then npm prune, then npm install (subsequent builds)

https://github.com/heroku/heroku-buildpack-nodejs#how-it-works

You do not want any surprises in your deploy process. Generally for production apps I will not use any version ranges (`^`, `~`, etc.) and I will [shrinkwrap my dependencies](https://docs.npmjs.com/cli/shrinkwrap). As such: for deploying a node app in production you probably don't want to be using npm update to build your slug.

Shrinkwrapping would be waay overkill for a simple gulp build process for a wordpress theme. People can definitely choose to shrinkwrap after they clone down this repo however. That way if they are working on a team or something they can have exactly reproducible builds.

### `npm install` vs `npm update`

Suppose you have:

```json
{
  "name": "nodetest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "lodash": "2.4.1"
  }
}
```

```bash
➜  nodetest  npm install
lodash@2.4.1 node_modules/lodash
```

Let's make a change to package.json

```diff
{
  "name": "nodetest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
-    "lodash": "2.4.1"
+    "lodash": "3.0.1"
  }
}
```

```bash
➜  nodetest  npm install
lodash@3.0.1 node_modules/lodash
```

So when you explicitly define a version: npm install will update the dependency.  Where npm install differs from npm update is this: 

If you have lodash 2.3.0 installed in your `node_modules` directory and you specified the version range such as `^2.3.0`:

- npm install will not do anything.
- npm update will update to 2.4.1

[`npm install` ignores version ranges.](http://stackoverflow.com/a/19824154/1585343). 

### This theme is not a node app

On a production node app your dependencies are very rarely version ranges. Because if you deploy your app and a dependency breaks, your app breaks and you just bought yourself downtime and everyone panics.

For a build process like Sage's: convenience and usability is going to win out over safety. I do not want people having to manually patch their theme's package.json to get backwards-compatible updates and bug-fixes. So we version ranges are okay. There will be edge-cases where our dependencies will break due to negligent package maintainers, but we should not be overly concerned with this. People who are paranoid about this should `npm shrinkwrap --dev` anyway.

As such: Travis needs to be testing the experience of someone who is cloning roots down for the first time. 

### Travis caching

http://docs.travis-ci.com/user/caching/#Fetching-and-storing-caches

> - We fetch the repo's cache on every build, including feature branches and pull requests.
> - There is one cache per branch and language version/compiler version/JDK version/Gemfile location/etc.
> - If a branch does not have its own cache yet, it will fetch the master branch cache.
> - Only modifications made to the cached directories from normal pushes are stored.

## Decision

So after deeply combing the npm repo and their issue tracker as well as running my own tests: `npm update` is incredibly strange. Try this: 

1. clone down the latest roots 8.0.0 in a temp directory or delete your `node_modules` existing directory
2. run "npm update"
3. it takes a while to update everything
4. run "npm update" again.
5. it takes just as long even though you just updated

It's actually faster on my machine to `rm -r node_modules` and run `npm install` than to run `npm update`. I am on npm 1.4.28 and Travis is on 1.4.28 as well. 

For scientific purposes let's go ahead and replace `npm install` with `npm update` and see what happens. I am curious and there is no clear documentation out there as to what is actually happening.

austinpray:
Ok so `npm update` is faster for cached builds. We went from like 7-8 minutes per build to 4 minutes.

austinpray:
With `npm update`

cache built? | total build time
---------|------------
no | 9 min 9 sec
yes | 8 min 10 sec

So not that big of a difference, but `npm update` is faster.

>It's actually faster on my machine to rm -r node_modules and run npm install than to run npm update. I am on npm 1.4.28 and Travis is on 1.4.28 as well.

I guess this behaviour is because `npm update` depends on network speed. Travis is inside a data center and not on my girlfriend's wifi so perhaps that's why `npm update` runs faster.

PR #1300, opened 2015-01-30
Uniform wp_link_pages between posts and pages

leopuleo:
wp_link_pages is now included in posts and pages with the same structure. Removed Bootstrap class.


retlehs:
nice catch, thanks. very tiny nitpick - you removed the blank line at the end of the file with this commit. do you have an editor that supports editorconfig?

see https://github.com/roots/roots/blob/8.0.0/.editorconfig#L11

leopuleo:
Yes, I'm using Sublime Text.
I see the Travis result, my fault! :grimacing:
I'm making a new commit, with the right end of the file!

leopuleo:
Oook :smiley: 
Now Travis is good!

PR #1301, opened 2015-01-31
[8.0.0] reconcile bower dependency versions with CDN url

austinpray:
- [x] memoize json reading
- [x] parse bower.json

## part 1

Memoize asset_path function

Previously roots would do a file operation each time asset_path is
called. So if you enqueued modernizr, jquery, and main.js it would read and parse dist/assets.json three times. Now it will only parse the json manifest one time per program execution.

Extract json reading to class that can be used to read bower.json

## part 2

fix https://github.com/roots/roots/issues/1286

### DIFF
--- a/lib/assets.php
+++ b/lib/assets.php
@@ -38,7 +38,6 @@ function asset_path($filename) {
   $dist_path = get_template_directory_uri() . '/dist/';
   $directory = dirname($filename) . '/';
   $file = basename($filename);
-
   static $manifest;
 
   if (empty($manifest)) {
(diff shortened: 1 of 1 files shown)


austinpray:
In the bower_map_to_cdn function can someone help me make it more fault tolerant? Idk how work with nested objects in php.

austinpray:
good to go?

lib/assets.php
```diff
@@ -17,24 +17,81 @@
...
+      $this->manifest = [];
+    }
+  }
+
+  public function get() {
+    return $this->manifest;
+  }
+
+  public function getPath($key = '', $default = null) {
+    $collection = $this->manifest;
```
swalkinshaw: line in between

lib/assets.php
```diff
@@ -17,24 +17,81 @@
...
+    if (is_null($key)) {
+      return $collection;
+    }
```
swalkinshaw: line in between

lib/assets.php
```diff
@@ -17,24 +17,81 @@
...
+    if (isset($collection[$key])) {
+      return $collection[$key];
+    }
```
swalkinshaw: line in between

lib/assets.php
```diff
@@ -17,24 +17,81 @@
...
+    foreach (explode('.', $key) as $segment) {
+      if (!isset($collection[$segment])) {
+        return $default;
+      } else {
+        $collection = $collection[$segment];
+      }
+    }
```
swalkinshaw: line in between

lib/assets.php
austinpray: prs acccepted

lib/assets.php
austinpray: prs acccepted

lib/assets.php
austinpray: prs acccepted

austinpray:
Anyone gonna fix the whitespace so I can merge? lol

PR #1306, opened 2015-02-04
Bumps asset-builder verison

austinpray:
use carrot to get latest minor version as well
ref https://github.com/austinpray/asset-builder/issues/14


swalkinshaw:
:shipit:

PR #1312, opened 2015-02-09
Fixing bootstrap-sass overrides

kalenjohnson:
Override directories weren't correct on a few of the bootstrap-sass assets.


retlehs:
oops. thanks @kalenjohnson

PR #1317, opened 2015-02-12
Remove /**/* glob from js by default

austinpray:
See https://discourse.roots.io/t/adding-none-bower-js-libs-to-gulp-manifest-json/2890?u=austin


austinpray:
`/**/*` will only work for extremely simple situations with disciplined developers who have js that looks like this:
good1

However I quickly imagine the average user's dep graph looks like this:
bad1

PR #1321, opened 2015-02-13
Migrate @Foxaii's nav-rejig for 8.0.0

retlehs:
what @Foxaii's nav-rejig does:

* fixes the default WordPress functionality where your blog/posts page is shown as **active** when viewing a single CPT post
* marks the proper menu item as active when necessary for CPT archives and CPTs

needs a code review and possible cleanup before merge

### DIFF
--- a/lib/nav.php
+++ b/lib/nav.php
@@ -2,6 +2,8 @@
 
 namespace Roots\Sage\Nav;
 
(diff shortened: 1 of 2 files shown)


retlehs:
move `root_relative_url`, `url_compare`, and `is_element_empty` to `lib/utils.php`?

Issue #1033, opened 2014-05-11
Better support for Custom Post Types


christianmagill:
I love the way Roots is organized, but why not include better support in the single.php and index.php templates for custom post types.

Perhaps grab the correct content with something like this...

```php
get_template_part('templates/content', get_post_type() != 'post' ? get_post_type() : get_post_format()  );
```

retlehs:
see #1322

PR #1322, opened 2015-02-13
Ref #1033 - better support for CPT templates

retlehs:
See #1033 

If your CPT is named `movie`:

* archive pages will attempt to load `templates/content-movie.php`
* single posts will attempt to load `templates/content-single-movie.php`

### DIFF
--- a/index.php
+++ b/index.php
@@ -8,7 +8,7 @@
 <?php endif; ?>
 
 <?php while (have_posts()) : the_post(); ?>
-  <?php get_template_part('templates/content', get_post_format()); ?>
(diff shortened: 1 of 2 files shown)


kalenjohnson:
Me likey

tiborp:
If we only had a 'like'button here ;-)

fgilio:
+1 I like it

ManuelRiveraC:
yupi!

rslnk:
good one!

cjmosure:
Nice +1

mattiapaiano:
+ 1

PR #1326, opened 2015-02-17
Fix missing `msgstr` in roots.pot

amekusa:
`msgid` and `msgstr` may have to be set.

http://www.gnu.org/software/gettext/manual/gettext.html#PO-Files


retlehs:
not accepting pulls to master right now. please submit this against 8.0.0

PR #1327, opened 2015-02-17
Fix missing "msgstr" in .POT

amekusa:
`msgid` and `msgstr` should always be in pairs.

http://www.gnu.org/software/gettext/manual/gettext.html#PO-Files


retlehs:
thank you

PR #1328, opened 2015-02-19
Re-organizing the styles directory

retlehs:
WIP, see commits + commit messages


retlehs:
my thoughts on removing `layouts/pages/_home.less`:

* we don't include a `front-page.php` or `home.php` template out of the box 
* activation is gone with 8.0.0, so we no longer add a home page
* i was thinking about adding a `_blog.less` file so `_home.less` wouldn't be so awkward/lonely, but being a starter theme, i don't think it should be an issue to remove this file. for simple sites, some people might just need to use `_pages.less`

retlehs:
image

schrapel:
If we are now importing the global bootstrap file from Bower. How do you recommend disabling parts of bootstrap we do not use?

JulienMelissas:
@schrapel - if you're talking about the CSS, I ran into this the other day. All I did was copied the contents of `bootstrap.less` (or .scss), into a new file called `_bootstrap.less` in my `assets/less/` folder, changed the paths to reflect the new paths, and added that file to `main.less`. Not sure if that's the recommended way from the team, but it worked fine for me.

If you're talking about the JS, you should delete the lines (don't comment them out, that breaks it) that you don't need in `bower.json` under the `bootstrap` section..

schrapel:
@JulienMelissas Yeah, I was talking about CSS. That was the file structure of 7.0 so I was wondering why they would remove _bootstrap.less if you have to do this process manually. 

I guess it is something to do with moving towards being framework agnostic.

JulienMelissas:
Yep, that move towards being framework agnostic is a great example of why we removed it. I think with 8.0 it's up to the developer a bit more to pick what tools/workflow they want. While us hard-core 7.0 users will need to do a tiny bit of extra work for every project, some things are way easier to do now too ;)

kalenjohnson:
Yeah, right now `bootstrap.less` is being included. You could either do what JulienMelissas suggests, or you could set up the overrides in `bower.json`:

https://github.com/roots/roots/blob/8.0.0/bower.json#L18-38

retlehs:
both bootstrap style and JS overrides should be changed in the `bower.json` as opposed to importing a new file

PR #1329, opened 2015-02-19
[WIP] Conditional tag check

swalkinshaw:
Simplify and generalize class since it had nothing to do directly with the sidebar. Now it can be used for other purposes as well.


retlehs:
note to self - update new docs

QWp6t:
I've actually been using the Sidebar class in this way for a while. I use it to determine whether to load a hero image (or other similar content). Oddly enough, it never really occurred to me to refactor it to make it more generic.

JulienMelissas:
This is great - I've also used the class for lots of things, (like heros). +1

swalkinshaw:
We had thought about it a while ago but mostly forgot about it. I think it was @Foxaii who had the idea.

I'm open to any better naming ideas as well although at least this one is explicit.

PR #1333, opened 2015-02-20
config.php clean up

retlehs:


retlehs:
minor clean up, also moves all the body class additions into one place in `lib/extras.php`

Issue #1340, opened 2015-02-23
wiredep is broken


retlehs:
broke after https://github.com/roots/roots/commit/2e62ebe8ffa25cce8d0ef4b4080c189b10dd16f5

PR #1343, opened 2015-02-23
Only save a wiredepped file if Sha1 differs

austinpray:
fixes https://github.com/roots/roots/issues/1340


retlehs:
#1340 will be fixed when:

* running `gulp` updates `main.less` when changes are made within the wiredep comments (eg. accidentally removing an import, or adding an import in there when it's not handled by wiredep)

* running `gulp` updates `main.less` when changes are made within `bower.json` that would affect the wiredep injection

austinpray:
Committed as separate commit so we can revert/reference later if needed

PR #1346, opened 2015-02-24
Add note about not manually editing wiredep injected deps

retlehs:


assets/styles/main.less
```diff
@@ -1,3 +1,4 @@
+// Automatically injected wiredep dependencies (never manually edit this block)
```
austinpray: Automatically injected bower dependencies via wiredep (never manually edit this block)

PR #1350, opened 2015-02-26
added localized string for 'Toggle navigation' in /templates/header.php

slackday:
A small change to keep strings in templates consistent. Output is the same as before as no translation exists yet.


retlehs:
nice catch, thank you

PR #1352, opened 2015-02-26
Handle $dist_path as var

emaildano:
Based on this thread: https://discourse.roots.io/t/sage-and-augmenting-my-workflow/2965/3

This sets up the opportunity to change the path of `dist/` for those needing to reference a different build.

### DIFF
--- a/lib/assets.php
+++ b/lib/assets.php
@@ -53,13 +53,14 @@ class JsonManifest {
 }
 
 function asset_path($filename) {
-  $dist_path = get_template_directory_uri() . '/dist/';
+  $dist_var = '/dist/';
+  $dist_path = get_template_directory_uri() . $dist_var;
   $directory = dirname($filename) . '/';
   $file = basename($filename);
   static $manifest;
 
   if (empty($manifest)) {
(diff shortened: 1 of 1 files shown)


austinpray:
wait what? where does $dist_path come from?

emaildano:
https://github.com/roots/sage/blob/master/lib/assets.php#L56

`$dist_path` was already in the file. How would one change the path if the build is something other than `dist/`?

austinpray:
Did you test this? It doesn't work. $dist_path is defined as such: `$dist_path = get_template_directory_uri() . '/dist/';`. there would need to be a separate variable `$dist_dir` or something. because the get_template_directory_uri will break stuff

emaildano:
@austinpray: I see what you mean. However it was working locally.

The change I made should work fine.

Personally my build will go to `dist-prod/` which will be checked into git and I'm adding a condition to target that in production. Here's just an example:

```
if (WP_ENV == 'production') {
  $dist_var = '/dist-prod/';
} else {
  $dist_var = '/dist';
}
```

austinpray:
Not super about a variable named $dist_var haha 
https://github.com/emaildano/roots/pull/1

PR #1355, opened 2015-02-27
travis: PHP 7.0 nightly added

TomasVotruba:
See http://docs.travis-ci.com/user/languages/php/#PHP-7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 sudo: false
 language: php
 php:
-- '5.6'
-- '5.5'
-- '5.4'
+  - nightly
+  - 5.6
+  - 5.5
+  - 5.4
+
+matrix:
+  allow_failures:
+    - php: nightly
 
 before_install:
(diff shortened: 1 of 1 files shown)


ntwb:
As `7.0` is only an alias of `nightly` and that PHP 7.0 has not yet been released I'd suggest you use `nightly` and allow these tests to fail.

For reference see WordPress' Core commit for the same: https://core.trac.wordpress.org/changeset/31558/

TomasVotruba:
Good points, I've implemented them.

retlehs:
not sure if we need this now, but please update the PR to use 2 spaces instead of 4

TomasVotruba:
@retlehs Fixed

QWp6t:
:+1:

It's a good idea to test future versions so we know what kind of issues to be cognizant of.

retlehs:
thanks yall

TomasVotruba:
Thanks!

PR #1363, opened 2015-03-03
Apply script_loader_src filter to jquery fallback

QWp6t:
This makes the fallback URL play nicely with Soil's relative URLs module.


QWp6t:
This could be refactored to work with #1352

retlehs:
:blue_heart:

PR #1383, opened 2015-03-13
Watch manifest.json

kalenjohnson:
Working on a project, when I want to modify manifest.json it helps the workflow if I don't need to run `gulp` every time I make a change.


austinpray:
:shipit:

kalenjohnson:
:sunglasses:

PR #1384, opened 2015-03-15
Remove mention of Google Analytics from the config

QWp6t:
~~Implements https://github.com/roots/soil/pull/58~~ 

this PR has been updated to remove the mention of GA from `lib/config.php`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### HEAD
+* Remove mention of Google Analytics from the config ([#1384](https://github.com/roots/sage/issues/1384))
+
 ### 8.1.1: March 31st, 2015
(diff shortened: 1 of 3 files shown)


retlehs:
since this will automatically output GA code (with `UA-XXXXXX` as the ID) if you update to the latest soil (after https://github.com/roots/soil/pull/58 is merged)... 

i think we could go ahead and just remove `add_theme_support('soil-google-analytics');` from `lib/config.php`, essentially removing any mention of GA from the theme

QWp6t:
Soil won't output anything related to GA unless `add_theme_support('soil-google-analytics', 'UA-XXXX-Y')` is added to the theme.

[If `$gaID` is falsey, Soil won't output anything.](https://github.com/roots/soil/blob/QWp6t-optionsGA/modules/google-analytics.php#L14)

[Default `$gaID` is an empty string](https://github.com/roots/soil/blob/QWp6t-optionsGA/modules/google-analytics.php#L35), which is falsey.

PR #1389, opened 2015-03-18
Changing SageNavWalker to NavWalker, using class directly in template

kalenjohnson:
Cleaning this up slightly, `SageNavWalker` still had the old style namespacing. Also since it's a class, we can reference it directly in the template without the last namespace connected to it.


swalkinshaw:
See https://github.com/roots/sage/pull/1390

PR #1396, opened 2015-03-23
Update README.md with folder structure section

kalenjohnson:
Adding a "Folder Structure" section, so people know what goes where, and that `dist` should not be touched manually.


kalenjohnson:
Sounds like we should add it to the docs on roots.io

Issue #1398, opened 2015-03-24
assets.json randomly munged when running gulp --production


natebeaty:
Just curious if anyone else has seen this. Running `gulp --production` will sometimes output an assets.json with an incomplete or mangled file list, such as this beauty:

```javascript
{
  "editor-style.css": "editor-style-03c28a52.css",
  "main.css": "main-432a583a.css"
}query.js": "jquery-e5581654.js",
  "svg4everybody.js": "svg4everybody-e9de7a9b.js",
  "main.js": "main-f66739a9.js"
}
```

It's maddeningly inconsistent—running  `gulp --production` repeatedly will sometimes work, then sometimes spit out a mangled file.

Before I dig further, just wanted to see if anyone else has run into this before.

austinpray:
Sounds like race condition city. I've never personally gotten this but I'm trying to reproduce it. What does your environment look like? Also `npm list --depth=0`

natebeaty:
Thanks for the quick response. Unfortunately I think this is going to be hard to reproduce. Though we have seen the issue on two separate computers in the office.

Not sure what all you need to know about my environment, happy to paste in anything that helps.

Our gulpfile.js: https://gist.github.com/natebeaty/d8b8a5b46a9e895d1e57 (there are some svg tasks added, but we saw the same behavior after replacing with default gulpfile.js)

manifest.json:

```
{
  "dependencies": {
    "main.js": {
      "files": [
        "scripts/main.js"
      ],
      "vendor": [
        "bower_components/history.js/scripts/bundled-uncompressed/html5/jquery.history.js",
        "bower_components/imagesloaded/imagesloaded.pkgd.js",
        "bower_components/fitvids/jquery.fitvids.js",
        "bower_components/slick-carousel/slick/slick.js",
        "bower_components/velocity/velocity.js",
        "bower_components/masonry/dist/masonry.pkgd.js"
      ],
      "main": true
    },
    "main.css": {
      "files": [
        "styles/main.scss"
      ],
      "main": true
    },
    "editor-style.css": {
      "files": [
        "styles/editor-style.scss"
      ]
    },
    "jquery.js": {
      "bower": ["jquery"]
    },
    "modernizr.js": {
      "bower": ["modernizr"]
    },
    "respond.js": {
      "bower": ["respond"]
    },
    "svg4everybody.js": {
      "bower": ["svg4everybody"]
    }
  },
  "config": {
    "devUrl": "sce.dev"
  }
}
```

```
[...]

PR #1401, opened 2015-03-30
Run styles and scripts in sequence

austinpray:
Prevents potential race condition where styles and scripts write to the
manifest at the same time

fixes https://github.com/roots/sage/issues/1398

Updates changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### HEAD
 * Update to Bootstrap 3.3.4
+* Fix `gulp --production` race condition ([#1398](https://github.com/roots/sage/issues/1398))
 
 ### 8.1.0: March 13th, 2015
 * Move HTML5 Boilerplate's Google Analytics snippet to Soil ([#1382](https://github.com/roots/sage/issues/1382))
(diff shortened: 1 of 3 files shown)


austinpray:
@retlehs pretty big bug. Should def be a release.

package.json
```diff
@@ -27,7 +27,6 @@
     "npm": ">=2.1.5"
   },
   "devDependencies": {
-    "pleeease": "3.2.0",
```
retlehs: think we need to keep this further up (or remove it if pleeease is now working as expected?)

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### HEAD
 * Update to Bootstrap 3.3.4
+* Fix `gulp --production` race condition. ([1398](https://github.com/roots/sage/issues/1398))
```
retlehs: this needs to match the rest of the changelog, so remove the period and add a `#` before the issue number :mag:

austinpray: Fixed

austinpray:
@retlehs I reverted all the changes to pleeease so this can get merged. Fixing the pleeease madness will have to be a separate PR.

PR #1416, opened 2015-04-10
Add support for asset-builder-cli

austinpray:
specify the manifest location here so CLI users do not have to type the `-m` or `--manifest` flag.

Simply:

`npm i -g asset-builder-cli@latest` and then run `assetbuilder` in theme root.

### DIFF
--- a/asset-builder.json
+++ b/asset-builder.json
@@ -1,0 +1,3 @@
+{
+  "manifest": "./assets/manifest.json"
+}


swalkinshaw:
Why not just make this the default in `asset-builder-cli`?

retlehs:
yeah, don't want to really add another file to the theme root

PR #1417, opened 2015-04-11
Remove $content_width

retlehs:


swalkinshaw:
Good riddance

kalenjohnson:
Happy days

austinpray:
Where were you when $content_width was removed? Never forget

JulienMelissas:
Bye `$content_width` - you won't be missed.

corradomatt:
Not that I mind, but why remove it now? Did I miss a discussion somewhere along the way?

JulienMelissas:
@corradomatt - there are a few reasons, and we did have a few conversations here and there (can't find records of it right now), but in general, it's because it's really not necessary, and can cause some issues and general confusion.

Here's a good resource on that: https://wycks.wordpress.com/2013/02/14/why-the-content_width-wordpress-global-kinda-sucks/

corradomatt:
Thanks for sharing that @JulienMelissas - makes sense to me.

PR #1418, opened 2015-04-11
Switch from yargs to minimist

austinpray:
For our simple usecase we do not need the extra functionality of yargs.
This removes 4 dependencies from our dependency graph.

https://github.com/bcoe/yargs
https://github.com/substack/minimist


swalkinshaw:
:+1:

PR #1430, opened 2015-04-19
Limit file patterns in gulp watch

QWp6t:
`gulp watch` was taking anywhere from 3 - 14 **minutes** (depending on disk activity, I assume) to start for me. This brings it down to a much more sane ~150ms.

The problem seems to be that `gulp watch` was walking all directories looking for PHP files, node_modules in particular. This limits it to `./`, `./lib/`, and `./templates/` directories.


swalkinshaw:
@austinpray

PR #1431, opened 2015-04-20
Update content-page.php

bdkabiruddin:
if i use only <?php the_content(); ?> on search result page showing whole content of page. sot it should <?php the_excerpt(); ?> with <?php the_title(); ?> like content.php file


JulienMelissas:
First of all, this build failed travis. There are some formatting issues (see https://travis-ci.org/roots/sage/jobs/59226032#L257).

Second, this might work for your particular case, but I think we'd like to keep the codebase as lean as possible. Just my opinion.

PR #1433, opened 2015-04-22
Clean before running a build

kevcenteno:
### Setup
1. Run `gulp --production`
2. Run `gulp watch`
3. Update a Less/Scss file

### Problem

The watch task catches the Less/Scss file update, but the browser still references the production file(s)

### Solution

1. Include `gulp:clean` in `gulp:build`
2. Set `gulp:build` as a dep task for `gulp:watch`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -215,7 +215,7 @@ gulp.task('clean', require('del').bind(null, [path.dist]));
 // `manifest.config.devUrl`. When a modification is made to an asset, run the
 // build step for that asset and inject the changes into the page.
 // See: http://www.browsersync.io
-gulp.task('watch', function() {
+gulp.task('watch', ['build'], function() {
   browserSync({
     files: [path.dist, '{lib,templates}/**/*.php', '*.php'],
     proxy: config.devUrl,
@@ -232,10 +232,10 @@ gulp.task('watch', function() {
 });
 
 // ### Build
-// `gulp build` - Run all the build tasks but don't clean up beforehand.
-// Generally you should be running `gulp` instead of `gulp build`.
+// `gulp build` - Run all the build tasks
 gulp.task('build', function(callback) {
-  runSequence('styles',
+  runSequence('clean',
+              'styles',
               'scripts',
               ['fonts', 'images'],
               callback);
@@ -256,6 +256,6 @@ gulp.task('wiredep', function() {
 
 // ### Gulp
 // `gulp` - Run a complete build. To compile for production run `gulp --production`.
-gulp.task('default', ['clean'], function() {
+gulp.task('default', function() {
   gulp.start('build');
 });


QWp6t:
I've seen this done with other projects as well. I'm indifferent about it. I don't see the harm, but it's not really something I've ever viewed as being in need of fixing.

Since @austinpray seems to handle the gulpfile stuff, I'd like to hear what he thinks about this.

kevcenteno:
The `clean` task adds about 16ms to the total build time (at least, on my machine)

joemaller:
If we're making the clean task a dependency of build, then the default task can be reduced to a simple alias:

```javascript
gulp.task('default', ['build']);
```
That lets us drop `gulp.start()` which is [frowned upon](https://github.com/gulpjs/gulp/issues/426#issuecomment-41208007) anyway.

austinpray:
>That lets us drop gulp.start() which is frowned upon anyway.

rekt

But to my affidavit I peeped that practice from [generator-gulp-webapp](https://github.com/yeoman/generator-gulp-webapp/blob/1c2b383310490522886b08c4fb205b5e424e44a4/app/templates/gulpfile.js#L161). That project has issues though, so not sure if it's a good example to follow anyway.

To give some background: the only reason we were doing it this way in the first place was to make sure clean was run before the build task. Now that we have runSequence we could pretty easily build clean into the default task.

The one thing I don't like about this is that it kills the ability to have incremental builds. My next major task for the gulpfile is introducing [incremental builds straight out of the 1980s](https://www.gnu.org/software/make/). I'm going to very quickly investigate some strategies available to us and then get back to ya'll. If it's going to take an overhaul I'll just merge this and make those changes later.

kevcenteno:
@austinpray Incremental build? Hmm...

On another note (this probably belongs in "Issues", but whatevs), we could probably save some CSS compile time by distinguishing production and dev builds.  One quick win would be to only run `minify` for production builds.  Another time suck is `writeToManifest`, which actually looks like two tasks in one (browsersync + rev); rev could probably be run only for production builds

joemaller:
@austinpray Im not a fan of generator-gulp-webapp, I feel like it tries to be too magical.

For incremental builds, [gulp-cached](https://www.npmjs.com/package/gulp-cached) might do a lot of what you want?

austinpray:
> Im not a fan of generator-gulp-webapp

Yeah me neither. I've used it on about a dozen projects and it has really become a burden. I usually just use [Make](https://github.com/clippy-io/clippy.io/blob/master/Makefile) these days on SPA projects. For WordPress/Magento type stuff I use the Sage gulpfile (obviously).

gulp-cached has changed a lot since I last used it, I need to investigate to see if it still has some issues. When I used it in the past it had issues with renaming files and stuff. I'll check it out.

@kevcenteno minifying all the time is intentional. You should be devving with the code that is going to production, not devving with dev code (unminified) and then crossing your fingers and pushing production code (minified). Incremental builds will speed up the process much more than disabling minification. Feel free to add a `--fast` flag to your build process that disables minification when the flag is set.

austinpray:
Sorry for leaving you hanging here. I have made headway on the incremental builds and this PR will be fundamentally incompatible with that feature.

PR #1448, opened 2015-05-03
Bump node dependencies

austinpray:
- [x] bump browsersync
- [x] bump gulp-sass


joemaller:
lol, I just did this too. BrowserSync also introduced a new `stream` method which simplifies the reload pipe a little, I'll PR that now.

PR #1450, opened 2015-05-03
Fix glyphicon font path

### LINKED ISSUE #1449: Compiled assets have incorrect file path
One a fresh install of 8.1.1 the file path to my bootstrap glyph icons is incorrect, here's an example of of what I'm getting in web inspector:

http://example.com/app/themes/sage/dist/styles/styles/fonts/bootstrap/glyphicons-halflings-regular.woff

I'm also getting the same problem with web fonts and image paths in my _global.less/scss files. I had add "../../" in front of the normal url ex:

src: url('../../../fonts/Aleo-Regular-webfont.eot');

usually the file path is src: url('../fonts/Aleo-Regular-webfont.eot');

fearlsgroove:
Fixes #1449

### DIFF
--- a/assets/styles/common/_variables.scss
+++ b/assets/styles/common/_variables.scss
@@ -1,3 +1,6 @@
+// Icon font path
+$icon-font-path: "../../../fonts/";
+
 // Grid settings
 $main-sm-columns:       12;
 $sidebar-sm-columns:    4;


retlehs:
thanks. a few things:

the path should be `../fonts/`, the comment should be updated to reflect that this is only for glyphicons, and the spacing should line up with the rest of the file:

```
// Glyphicons font path
$icon-font-path:        "../fonts/";
```

austinpray:
https://github.com/roots/sage/pull/1452

austinpray:
@fearlsgroove please rebase your commit to include the suggestions in https://github.com/roots/sage/pull/1450#issuecomment-98698878

PR #1451, opened 2015-05-03
Update page.php for a better search results list

leopuleo:
Thanks to 
`<?php get_template_part('templates/content', get_post_type() != 'post' ? get_post_type() : get_post_format()); ?>` 
it's amazing to manage different post type archive / taxonomy term page without crazy files duplication.

Unfortunately, this breaks search results list because `content-page.php` is in templates folder.

screenshot

I propose to remove `content-page.php` and move the code in `page.php`.
I've noticed this after PR #1431


retlehs:
i'm not sure what the best solution is for this, but i don't really like the idea of removing `content-page.php`

i also don't like the idea of creating `search.php` + `content-search.php` files, or adding `is_search` conditionals anywhere

:weary:

leopuleo:
I agree with you @retlehs, I chose the most drastic solution...

What about a simple file re-naming:
`content-page.php -> page-content.php`

It's similar to the actual situation, but the name can confuse.
By the other hand, it's coherent with `page-header.php`

It's hard...any solution can be right or wrong :smiley:

retlehs:
after sleeping on this and chatting with @Foxaii, this is what we're gonna roll with: #1459 

thanks a lot for bringing this up!

leopuleo:
Very well, sleeping always brings a solution :wink:

PR #1452, opened 2015-05-04
Fix incorrect file path in compiled assets

austinpray:
ref https://discourse.roots.io/t/compiled-assets-have-incorrect-file-path/3672/8

closes https://github.com/roots/sage/issues/1449

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -103,7 +103,10 @@ var cssTasks = function(filename) {
         'opera 12'
       ]
     })
-  .pipe(minifyCss)
+    .pipe(minifyCss, {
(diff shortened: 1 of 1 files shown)


swalkinshaw:
:100:

ironprogrammer:
Downgrading to gulp-minify-css 1.1.0 can also work around this issue, from personal experience and as noted here: https://github.com/jonathanepollack/gulp-minify-css/issues/95.

To do this, in package.json set the dependency for gulp-minify-css to "1.1.0", no caret. Then run `npm update` in your theme directory.

PR #1457, opened 2015-05-06
update BrowserSync

joemaller:
This PR switches to BrowserSync's recommended "[post 2.0.0 syntax](http://www.browsersync.io/docs/api)" and uses the cleaner [`stream`](http://www.browsersync.io/docs/api/#api-stream) method for reloading. The stream method offers glob filtering, so we are able to simplify the reload pipe in `writeToManifest` a bit. 

This is compatible with #1448.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -196,7 +196,8 @@ gulp.task('scripts', ['jshint'], function() {
 gulp.task('fonts', function() {
   return gulp.src(globs.fonts)
     .pipe(flatten())
-    .pipe(gulp.dest(path.dist + 'fonts'));
+    .pipe(gulp.dest(path.dist + 'fonts'))
+    .pipe(browserSync.stream());
 });
 
 // ### Images
@@ -208,7 +209,8 @@ gulp.task('images', function() {
       interlaced: true,
       svgoPlugins: [{removeUnknownsAndDefaults: false}]
     }))
-    .pipe(gulp.dest(path.dist + 'images'));
+    .pipe(gulp.dest(path.dist + 'images'))
+    .pipe(browserSync.stream());
 });
 
 // ### JSHint
@@ -234,7 +236,7 @@ gulp.task('clean', require('del').bind(null, [path.dist]));
 // See: http://www.browsersync.io
 gulp.task('watch', function() {
   browserSync.init({
-    files: [path.dist, '{lib,templates}/**/*.php', '*.php'],
+    files: ['{lib,templates}/**/*.php', '*.php'],
     proxy: config.devUrl,
     snippetOptions: {
       whitelist: ['/wp-admin/admin-ajax.php'],
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,7 +1,7 @@
 // ## Globals
 var argv         = require('minimist')(process.argv.slice(2));
 var autoprefixer = require('gulp-autoprefixer');
-var browserSync  = require('browser-sync');
+var browserSync  = require('browser-sync').create();
 var changed      = require('gulp-changed');
 var concat       = require('gulp-concat');
 var flatten      = require('gulp-flatten');
@@ -143,9 +143,7 @@ var jsTasks = function(filename) {
 var writeToManifest = function(directory) {
   return lazypipe()
     .pipe(gulp.dest, path.dist + directory)
-    .pipe(function() {
-      return gulpif('**/*.{js,css}', browserSync.reload({stream:true}));
-    })
+    .pipe(browserSync.stream, {match: '**/*.{js,css}'})
     .pipe(rev.manifest, revManifest, {
       base: path.dist,
       merge: true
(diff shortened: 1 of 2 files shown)


austinpray:
This reloads javascript twice due to the fact that the .map file gets through. Have the pattern exclude .map files

screenshot 2015-05-05 23 23 50

michaelbeil:
lol @austinpray

austinpray:
@michaelbeil :eyes:

joemaller:
I was just about to post a question about that over on Discourse... I think it's actually a problem over here: 
https://github.com/roots/sage/blob/master/gulpfile.js#L239

If you remove `path.dist` from that line, it works correctly and injects changes without a full reload. I'm currently checking into how much else that would affect.

retlehs:
please stop making unnecessary comments on github, especially emoji ones. lots of us get email notifications for each comment. thanks!

joemaller:
The extra 'File changed' notice likely results from a change in BrowserSync's matcher. (This would then affect  #1448 as well.) As noted above, [line 239](https://github.com/roots/sage/blob/master/gulpfile.js#L239) now seems be recursively reading `path.dist`.

Checking against master with only BrowserSync updated, any file changed inside the dist directory is now triggering a full reload. 

Rolling back to BrowserSync v2.5, *nothing* happens when touching `dist` or changing files inside it?! This means changing images or fonts do not triggering a reload either. :confused:

gulpfile.js
```diff
@@ -143,9 +143,7 @@ var jsTasks = function(filename) {
 var writeToManifest = function(directory) {
   return lazypipe()
     .pipe(gulp.dest, path.dist + directory)
-    .pipe(function() {
-      return gulpif('**/*.{js,css}', browserSync.reload({stream:true}));
-    })
+    .pipe(browserSync.stream, {match: '**/*.{js,css}'})
```
austinpray: Actually just remove this. With the new version the path.dist in files is enough to make things work.

joemaller:
I'm testing two solutions:

1. Change the BrowserSync files array to include `path.dist + '{fonts,images}/**/*'`
2. Remove `path.dist` from the BrowserSync files array, then explicitly trigger a reload from the images task and fonts task streams with something like `.pipe(browserSync.stream())`

I think the second option is probably best.

austinpray:
Option 1 works great, what's wrong with that? I use option 1 more frequently than all this stream stuff.

gulpfile.js
joemaller: seriously? I'll try that too.

austinpray: Yep it reloads images and such too.

joemaller: I don't think that really works. I'm seeing it picking up every changed file in `dist`, including source maps. The additional files mess up BrowserSync's asset-injection and force a full page reload. With a lot of files, they can also send multiple reload triggers. 

pastedgraphic-2

austinpray: JavaScript files should be a full page reload. CSS injecting is working properly over here after omitting the inline reload.

joemaller: Can you compare my [bs_files-only](https://github.com/joemaller/sage/tree/bs_files-only) branch with the code you're seeing running? I'm only seeing full reloads from that, no CSS injection and a ton of changed file notices.

joemaller:
Using BrowserSync's stream support means there's no need to watch output files for changes-- the output files' contents are already being piped through the stream. This saves the overhead of additional file watches and reading back the same file we just wrote from the task. It doesn't make sense to watch a stream's output file for changes when the stream itself is available. 

I think this should be nailed down now.

Two other branches were pushed to my fork to isolate the solutions described above for testing:  [bs_path-dist-mod](https://github.com/joemaller/sage/tree/bs_path-dist-mod)   and  [bs_files-only](https://github.com/joemaller/sage/tree/bs_files-only).

austinpray:
I just tested out https://github.com/joemaller/sage/tree/bs_path-dist-mod

Yeah let's definitely do that. Works perfectly. Also you are right: the less files we watch the better. Especially on Windows.

gulpfile.js
```diff
@@ -198,7 +196,8 @@ gulp.task('scripts', ['jshint'], function() {
 gulp.task('fonts', function() {
   return gulp.src(globs.fonts)
     .pipe(flatten())
-    .pipe(gulp.dest(path.dist + 'fonts'));
+    .pipe(gulp.dest(path.dist + 'fonts'))
+    .pipe(browserSync.stream());
```
austinpray: :+1:

gulpfile.js
```diff
@@ -210,7 +209,8 @@ gulp.task('images', function() {
       interlaced: true,
       svgoPlugins: [{removeUnknownsAndDefaults: false}]
     }))
-    .pipe(gulp.dest(path.dist + 'images'));
+    .pipe(gulp.dest(path.dist + 'images'))
+    .pipe(browserSync.stream());
```
austinpray: :+1:

PR #1460, opened 2015-05-09
Updating Google CDN info/instructions

JulienMelissas:
Open to comments or other wording ideas on this. We could remove completely if we want now that Soil needs to be installed for anything to work anyways...


JulienMelissas:
Opening new PR to remove - closing this one out.

PR #1462, opened 2015-05-09
add missing navbar-nav class

kevinlisota:
When the navwalker was removed, it looks like the navbar-nav class got removed as well.

That class is still necessary for proper rendering of a Boostrap navbar.


retlehs:
it was removed on purpose

PR #1422, opened 2015-04-12
Move jQuery CDN feature to Soil

retlehs:
See https://github.com/roots/soil/pull/64

Closes #1408

PR #1468, opened 2015-05-14
Remove unnecessary references to jQuery

retlehs:
no longer needed after #1422

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -17,9 +17,6 @@
         "styles/editor-style.scss"
       ]
     },
-    "jquery.js": {
-      "bower": ["jquery"]
-    },
     "modernizr.js": {
       "bower": ["modernizr"]
     }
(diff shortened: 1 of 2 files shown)


JulienMelissas:
Can't believe I forgot about this while I was looking through everything - good catch!

JasonTheAdams:
So if it's not being loaded via bower, then how is jquery being loaded? Or did you just opt to let Wordpress do it the default way?

austinpray:
I need to do some quick research on how to tell bower a dependency is included externally not via bower. Bower needs to know what version of jQuery we are using so people are prevented from installing  packages that require conflicting versions of jQuery.

@JasonTheAdams yes we use the default WordPress jQuery version.

retlehs:
merging this for now. @austinpray you can follow up with another PR

QWp6t:
@austinpray 

bower/bower#1394
bower/bower#927

PR #1469, opened 2015-05-15
change hooks to anonymous functions

JasonTheAdams:
It's unnecessary overhead to create functions that will only ever be used in hooks. This cleans up the hooks that could be changed to anonymous functions.

It didn't fit this PR, but I also recommend moving the utils.php hook to extras.php or something. It's now an unused namespace and doesn't really fit there, in my opinion. I'd imagine general purpose, utility functions to live there.

### DIFF
--- a/lib/assets.php
+++ b/lib/assets.php
@@ -65,7 +65,7 @@ function asset_path($filename) {
   }
 }
 
-function assets() {
+add_action('wp_enqueue_scripts', function() {
   wp_enqueue_style('sage_css', asset_path('styles/main.css'), false, null);
 
   if (is_single() && comments_open() && get_option('thread_comments')) {
@@ -74,5 +74,4 @@ function assets() {
 
   wp_enqueue_script('modernizr', asset_path('scripts/modernizr.js'), [], null, true);
   wp_enqueue_script('sage_js', asset_path('scripts/main.js'), ['jquery'], null, true);
-}
-add_action('wp_enqueue_scripts', __NAMESPACE__ . '\\assets', 100);
+});
--- a/lib/assets.php
+++ b/lib/assets.php
@@ -65,7 +65,7 @@ function asset_path($filename) {
   }
 }
 
-add_action('wp_enqueue_scripts', function() {
+add_action('wp_enqueue_scripts', function () {
   wp_enqueue_style('sage_css', asset_path('styles/main.css'), false, null);
 
   if (is_single() && comments_open() && get_option('thread_comments')) {
--- a/lib/extras.php
+++ b/lib/extras.php
@@ -7,7 +7,7 @@ use Roots\Sage\Config;
 /**
  * Add <body> classes
  */
-function body_class($classes) {
+add_filter('body_class', function ($classes) {
   // Add page slug if it doesn't exist
   if (is_single() || is_page() && !is_front_page()) {
     if (!in_array(basename(get_permalink()), $classes)) {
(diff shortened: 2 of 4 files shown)


JulienMelissas:
IMHO this is a great use of anonymous functions. :+1:

jasongottschalk:
How would such a function be unhooked?

JasonTheAdams:
There's been discussion on unhooking anonymous functions, and some folks have submitted [functions that are capable of removing them](http://wordpress.stackexchange.com/questions/137688/remove-actions-filters-added-via-anonymous-functions).

Still, I think the safer answer is to never use an anonymous function in a scenario wherein you could imagine someone wanting to unhook it. Plugins, as an example, should almost always use a named function, as you know you're code is becoming a part of something bigger.

Sage, on the other hand, is a starting point. If someone wanted to unhook one of the base functions, they'd either modify it or just delete the code altogether. In general, I always use anonymous functions for my theme hooks, because I'm confident no one will be unhooking them and it makes for cleaner code.

Hope that helps!

kalenjohnson:
@jasongottschalk I'm a little torn on this for that reason that it can't be unhooked. However, this is a starter theme and generally the end result, so unhooking is not nearly as important as in a plugin.

Still, the ability to unhook is a core attribute of WP. So I think we would want to make sure none of these would need to be unhooked for any reason, even say in the theme if it would need to be unhooked for a specific page or area.

JulienMelissas:
> even say in the theme if it would need to be unhooked for a specific page or area

Ahhhhh... That's an important point to mention.
After looking at this a little further I thought "well how do you unhook it" - and then I realized this is a starter theme, so you'd just delete the code you don't like or aren't using.

But now that @kalenjohnson brings up the fact that you might have to unhook this in your own theme, I can think of times where I might need that. I know that Sage tends to stray from default WP rules in a few small cases, but I'm not sure the sake of a little bit cleaner code (it is nicer to look at) is worth breaking functionality that is inherent to WordPress.

JasonTheAdams:
I can definitely appreciate all the thoughts. Personally, I can't think of a single time that I've unhooked one of my own functions. I've conditionally hooked things, but generally if it's a function I've developed, I'll have the function return null/default within the hook, rather than unhook it at a later time.

For me, I'd say especially the few things here that I've applied the anonymous functions to are pretty safe. I would sooner edit the contents of the functions or delete it altogether than unhook it. Still, I see the call you're making. Up to you. :)

mAAdhaTTah:
I was going to say, you don't really need to unhook your own functions so much as you would hook them conditionally. Or at least, you wouldn't need to unhook these functions. If you have functions you need to unhook, then make them named functions.

JulienMelissas:
^ for sure. Just was putting it out as a possible senario, and we do have to remember the system we're working within. Have I ever had to unhook anything I also wrote in the theme? Nope. Would love to hear some other opinions on this. Again, I will say it does _look_ better for sure.

christianmagill:
Functions that use a specific hook or filter can have multiple purposes. Having a proper function with a function name provides good self-documentation.

QWp6t:
:-1: 

There is no "unnecessary overhead" being alleviated by this PR as the anonymous function still exists in the global scope as part of `$wp_filter`.

JasonTheAdams:
As I've given this more thought, I think it is more cosmetic than anything. Personally, I like them for hooks as I'm absolutely confident there will never be a naming conflict, it's easy to look at and understand, and I don't have to worry about namespaces. That being said, since Sage puts absolutely everything in a namespace, it's really less of an issue. Namespaces, in my opinion, quickly become cumbersome (Java, anyone?), and so I try not to use them for their own sake.

This PR definitely shows the fact that this is a matter of preference. I'm tempted to say we should just close it, as it raises concern of being able to unhook the functions. While I never do this in my code, that doesn't mean others don't, and I'd rather not frustrate their workflow. @QWp6t is right, and I was wrong, that there really isn't any overhead benefit. I don't know if PHP handles anonymous functions any differently internally, but any potential benefit would likely be infinitesimal.

retlehs:
thanks everybody, going to pass on this - good discussion!

Issue #1470, opened 2015-05-15
Start/Close HTML in wrapper?


JasonTheAdams:
One thing that confuses me is why the HTML tag is opened in the header template, but closed in the wrapper? Why not just open/close the HTML tag in the wrapper? It's a small thing but it's always nicer when a tag opens and closes in the same file.

retlehs:
was thinking we could probably just get rid of `head.php`

JulienMelissas:
Not sure how I feel about getting rid of `head.php` all-together (in case there are projects with lots of `base-*.php` files but it's a great idea to move the opening `<html>` tag into base.php.

JasonTheAdams:
@JulienMelissas The issue here is that you're either trying to anticipate people with a lot of base template or head templates. My guess, though, is that there wouldn't be many with a unique combination of both, and since the base idea is generally liked, I'd be open for simply having having a more complete base.php.

People can always make room for a head template if they'd like

What do you think?

retlehs:
i'm pretty confident that there should never be a case where there are lots of `base.php` files.

moving `head.php` directly into `base.php` also seems like a safe change to make

JulienMelissas:
I don't think that there would ever be more than one head template, and I do get that there really should be minimal base.php templates. But I've been in projects where there were 3+ base.php templates, and in that case, yes, I could have made a `head.php` and just linked to that so that there was only one partial for that code.

guess that's a :+1: to this one

JulienMelissas:
@JasonTheAdams you trying to make a PR or you want one of us to do it?

JasonTheAdams:
Beat me to it. :)

JulienMelissas:
lol sorry we were just talking about it a bit internally and realized we wanted to keep the `head.php` partial and I figured it would be quicker for me to submit a PR than to explain it :wink: 

thanks again for pointing this out!!!

JasonTheAdams:
Certainly! Cheers!

PR #1471, opened 2015-05-15
Moved opening html tag to base.php

JulienMelissas:
Still using the `head.php` partial at this point.


JulienMelissas:
This should close out https://github.com/roots/sage/issues/1470

kalenjohnson:
Can't believe I missed this all these years

:shipit:

PR #1472, opened 2015-05-16
Keep SVG element IDs during gulp images task

eballeste:
Includes a cleanupIDs set to false option within the gulp images task in order to preserve the ID's set for internal SVG elements.


austinpray:
@retlehs good to go

retlehs:
@austinpray good 2 go tacos? yes, i would like some please

thanks @eballeste!

PR #1477, opened 2015-05-26
Modify wrapper to conditionally show sidebar-*.php templates for custom post types

vonkanehoffen:
Not sure if this is a particularly elegant way to do it, but it's helped me so thought I'd share! 

Being able to specify different sidebars for different post types cuts down on the need for duplicate `base-*.php` files. Obviously these sidebars have to be declared in `widgets_init()` too with this solution, but still.... an idea :-)

### DIFF
--- a/base.php
+++ b/base.php
@@ -25,7 +25,7 @@ use Roots\Sage\Wrapper;
         </main><!-- /.main -->
         <?php if (Config\display_sidebar()) : ?>
           <aside class="sidebar" role="complementary">
-            <?php include Wrapper\sidebar_path(); ?>
+            <?php include Wrapper\sidebar_path(Config\custom_sidebar()); ?>
           </aside><!-- /.sidebar -->
         <?php endif; ?>
       </div><!-- /.content -->
--- a/lib/config.php
+++ b/lib/config.php
@@ -63,3 +63,21 @@ function display_sidebar() {
 
   return $display;
 }
+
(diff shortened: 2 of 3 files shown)


Foxaii:
This should be very easy to accomplish in the same fashion we can with the base templates (see the modified and untested code below!). Doing it this way you won't need to manually create the array of sidebar post types; if the template for the post type is there, it will be used.

I also think that the "routing" of sidebars and base templates should be kept to a bare minimum in the theme as the WordPress template hierarchy, which is all builds upon, is complicated enough.

```php
add_filter('sage/wrap_sidebar', __NAMESPACE__ . '\\sage_wrap_sidebar_cpts'); // Add our function to the sage/wrap_sidebar filter

function sage_wrap_sidebar_cpts($templates) {
  $cpt = get_post_type(); // Get the current post type 
  if ($cpt) {
    array_unshift($templates, 'templates/sidebar-' . $cpt . '.php'); // Shift the template to the front of the array
  }
  return $templates; // Return our modified array with templates/sidebar-$cpt.php at the front of the queue
}
```

swalkinshaw:
As @Foxaii showed this is pretty easy to achieve via filters. So this isn't something we'd want to include by default.

vonkanehoffen:
Ah that way of doing it is loads better @Foxaii ! Hadn't thought of that! :smile:

PR #1482, opened 2015-06-02
Adjust comments on ConditionalTagCheck to match actual functionality.

Psykar:
The check returns false, if any of it's tags are true, not vise versa.

To be perfectly honest this behaviour seems opposite of what I'd expect (should return true if any of them are true) but it would be a breaking change, so for now just update the comments to match.

### DIFF
--- a/lib/conditional-tag-check.php
+++ b/lib/conditional-tag-check.php
@@ -4,7 +4,7 @@ namespace Roots\Sage;
 
 /**
  * Utility class which takes an array of conditional tags (or any function which returns a boolean)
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Yeah, it's confusing. This class was sidebar specific before so it returned pages you **don't** want the sidebar on. However as it's become more generic it makes less sense.

We could just negate the result here: https://github.com/roots/sage/blob/fd8d9accb7c51b40113b862510ed4bf4c19fecef/lib/config.php#L61

swalkinshaw:
Merging this for now but further updates are welcome.

Issue #1480, opened 2015-05-28
Problems debugging javascript


aaemnnosttv:
While working with a Sage project, `gulp; gulp watch`

`uglify` is stripping out any `debugger;` statements (not to mention `jshint` complains about it too).

I can add `"debug": true` to `.jshintrc` but that obviously isn't great for production use.

It seems we need to be able to pass some different options to `jshint` and `uglify` if we're not running `gulp --production`.  Perhaps forego `uglify` at all if not production?

austinpray:
>Perhaps forego uglify at all if not production?

That's a negative Ghost Rider. You want to be running when goes into production on dev. I do not cross my fingers and hope it works when I deploy to staging and production.

>It seems we need to be able to pass some different options to jshint and uglify if we're not running gulp --production.

Yeah let's only enable `drop_debugger` for `--production` runs. I'll get do it unless you say you want to do it.

aaemnnosttv:
Yeah, I think that's best.  Is there a way to do the same for `jshint`?

austinpray:
Well I guess you want it to stay in JSHint so you know the debug statement should be removed at some point right? JShint should only fail the build if it's `--production`. In dev it's a "by the way".

austinpray:
Yeah so JSHint was failing the task always. Now it only fails the build when you run scripts through `--production`.

PR #1487, opened 2015-06-04
Allow `debugger` statements in development JS

austinpray:
fixes https://github.com/roots/sage/issues/1480

cc @aaemnnosttv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### HEAD
 * Add search templates ([#1459](https://github.com/roots/sage/issues/1459))
+* Allow `debugger` statements in development JavaScript ([#1487](https://github.com/roots/sage/issues/1487))
 
 ### 8.2.1: May 7th, 2015
 * Update BrowserSync ([#1457](https://github.com/roots/sage/issues/1457))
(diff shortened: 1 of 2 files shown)


austinpray:
I need some help with the decision to fail the JSHint task. 

When JSHint fails:
- what should happen when in development mode?
- what should happen when in `--production` mode?

I was thinking that JSHint should give "by the way" warnings in development but fail production builds. But I'm unsure if this is the way to go. Really would appreciate feedback.

On one hand it makes sure no bad/dangerous code makes it to production. On the other: if you ignore the development JSHint warnings you will be unable to deploy until you pay the JSHint debt.

Foxaii:
The "by the way" warnings are a good idea. One long list of all the changes that need attention is better than just the first fault that has caused the build to fail.

I agree with your thinking that it should be a fail in `--production` as well. Perhaps document that they need to run in development mode to list the fixes required.

JulienMelissas:
I agree with @Foxaii - I think it's best if during development it just provides warnings, and during production it should fail. Documentation is a good think to bring up too, thanks for mentioning that!

austinpray:
>Perhaps document that they need to run in development mode to list the fixes required.

@Foxaii: in both cases it lists all the errors you need to fix. The only difference is that the production build will fail to compile.

Foxaii:
Nice to know but that doesn't change my agreeing with you, @austinpray.

austinpray:
ping @aaemnnosttv, curious to see what you think about this.

aaemnnosttv:
@austinpray I think it's a great solution.  Great job :+1:

PR #1488, opened 2015-06-07
Added the sage favicon.ico

carl-alberto:
Users may need to update their own favicon in the assets/images directory


retlehs:
a favicon.ico in the web root is better, imo - don't think this should be in master. let users decide what they wanna do.

PR #1492, opened 2015-06-11
Support for svg in asset pipeline

viktorbergehall:
As discussed at https://discourse.roots.io/t/adding-svg-support-to-the-asset-pipeline

The approach of this solution is largely inspired of this css-tricks article https://css-tricks.com/svg-sprites-use-better-icon-fonts/

### Overview:
- Svg files are stored in assets/svg
- Those files are minified and injected to a file called templates/content-svg.php
- content-svg.php gets included in base.php
- For better browser support a png fallback is created for every svg file

### Usage:
The svg files are converted to symbol elements in content-svg.php. Every symbol get the id from the original file name, ie menu.svg becomes <symbol id="menu" ...
To reference a symbol you simply use:

```html
<svg>
  <use xlink:href="#"></use>
</svg>
```

In order for this to work, the content-svg.php file have to be included. I have included it in the base.php file.

### Improvements:
- I didn't really get the hang of the globs variables in the Gulpfile. I guess you could store svg paths in there for better consistency.
- Better automated workflow for adding png fallbacks

### DIFF
--- a/assets/svg/content-svg.php
+++ b/assets/svg/content-svg.php
@@ -1,0 +1,3 @@
+<div style="height: 0; width: 0; position: absolute; visibility: hidden">
+  <!-- inject:svg --><!-- endinject -->
+</div>
--- a/base.php
+++ b/base.php
@@ -16,6 +16,7 @@ use Roots\Sage\Wrapper;
     <![endif]-->
     <?php
       do_action('get_header');
+      get_template_part('templates/content-svg.php');
       get_template_part('templates/header');
     ?>
     <div class="wrap container" role="document">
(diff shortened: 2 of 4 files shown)


AdamWills:
I use the svg.php myself, so this is interesting... looks like it needs some work in terms of code clean, but like the idea of this task.

retlehs:
thanks for the PR, but going to pass on including this by default. it's definitely useful, so [sharing your diff](https://github.com/roots/sage/pull/1492/files) on the discourse [thread](https://discourse.roots.io/t/adding-svg-support-to-the-asset-pipeline) would be helpful to others who come across it.

lekkerduidelijk:
Just a big thank you to @viktorbergehall for your work. This still works for Sage 8.5.0. 👏

PR #1494, opened 2015-06-12
Remove ConditonalTagCheck class

QWp6t:
Note that `config.php` now encourages users to use the following syntax:

```php
      is_404(),
      is_front_page(),
      is_page_template('template-custom.php'),
```

The old method of using an array of callbacks is still supported if for whatever reason someone still wants to do that, but I don't see the point in using that method. 

Honestly I think we could just remove the callback support all together and replace `display_sidebar()` with: 
```diff
   static $display;
 
   if (!isset($display)) {
-    $display = !Utils\conditional_reduction([
+    $display = !in_array(true, [
       /**
        * Any of these conditional tags that return true won't show the sidebar.
        * You can also specify your own callbacks as long as they each return a boolean.
        *
        * Examples:
        *
        * is_single()
        * is_archive()
        * is_page('about-me')
        * is_tax(['flavor', 'mild'])
        * is_page_template('about.php')
        * is_post_type_archive(['foo', 'bar', 'baz'])
        * \MyNamespace\MyClass::MyMethod(['my', 'args'])
        *
        */
       is_404(),
       is_front_page(),
       is_page_template('template-custom.php'),
     ]);
   }
 
   return apply_filters('sage/display_sidebar', $display);
 }
```


swalkinshaw:
I'm fine with removing callback support. Anyone can just do something like:

```php
$display = $display || my_custom_check();
```

QWp6t:
So what's the verdict? Are we dropping it entirely?

It was kinda fun refactoring the function, but I won't miss it if it's gone. I don't think anyone else will either. I can't imagine any users being terribly attached to the awkward syntax with the arrays.

swalkinshaw:
@QWp6t drop it

QWp6t:
:ok_hand:

QWp6t:
I added a couple other changes.

 * `$display` is no longer a static variable. Developers can modify the function as needed if they need to cache results.
 * Examples have been removed. Since the function itself has been simplified, I think the large block of docs would overwhelm new developers, i.e., make them think this is more complicated than it actually is.

I believe these changes are more consistent with our efforts to declutter and minimize Sage.

JulienMelissas:
So much simpler and easy to use. I'm all for this one. Thanks @QWp6t :heart_decoration:

Foxaii:
I don't think we should drop the static nature of the variable. Some conditionals can vary their return depending on where in the page load process (pre loop, in the loop, post loop) they are, so this could lead to some inconsistencies.

QWp6t:
That's initially why I added the static variable back in #1196, but you could just as easily hook into `sage/display_sidebar` filter to accomplish the same thing or modify the function to set your own static variable.

As I mentioned above, I think the change is consistent with our recent moves toward decluttering and minimizing Sage. By default we don't need it to be static, and I think most users don't need that either.

But obviously I'll add the static variable back if everyone agrees it should be there. lol

swalkinshaw:
Just a note that we're going to require some docs updates after this: https://roots.io/sage/docs/theme-sidebar/

Foxaii:
I'm still very much against dropping the static variable.

The only benefit of removing it is it won't be there. Keeping it improves performance by not unnecessarily repeating a bunch of conditional calls and guarantees the function returns consistently, which is a real issue considering some conditionals vary throughout the load process.

Dropping one keyword makes it less performant and less reliable. That's not the kind of decluttering we should be doing.

QWp6t:
@Foxaii :watermelon:

PR #1496, opened 2015-06-14
Fix source map paths

martinvd:
This sets the gulp-sourcemaps `sourceRoot` option, now pointing to the right locations of the source maps. Since it was not set, it was pointing to the default [`/source/`](https://github.com/floridoo/gulp-sourcemaps/blob/master/index.js#L191) folder, which is not the right folder.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -111,7 +111,9 @@ var cssTasks = function(filename) {
       return gulpif(enabled.rev, rev());
     })
     .pipe(function() {
-      return gulpif(enabled.maps, sourcemaps.write('.'));
+      return gulpif(enabled.maps, sourcemaps.write('.', {
+        sourceRoot: 'assets/styles/'
+      }));
     })();
 };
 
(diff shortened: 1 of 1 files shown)


austinpray:
I have some users on the forums testing this out. I'll test it out personally here in a second.

In the mean time: @swalkinshaw is probably going to make you squash your two commits into one. :cop:

martinvd:
Thanks for your reply :+1: Sorry that I didn't do an interactive rebase btw!

martinvd:
I just squashed the two commits into one.

austinpray:
@retlehs good to go

austinpray:
@martinvd What do you think about this? https://discourse.roots.io/t/error-loading-source-scss-files/4060/5?u=austin

martinvd:
That's interesting. I'll have a look into this asap and will try to reproduce it as well and then respond in the Roots Discourse issue.

martinvd:
@austinpray That's interesting. I'll have a look into this asap and will try to reproduce it as well and then respond in the Roots Discourse issue.

austinpray:
Thanks bud!

martinvd:
np :+1:

PR #1507, opened 2015-07-01
Removing PHP short tags as per WP theme review standards

basteln3rk:
PHP short tags, such as ``<?=$variable?>``, are disabled by many web hosts and therefore discouraged according to the WP [theme review standards](https://make.wordpress.org/themes/handbook/review/required/theme-check-plugin/)

Removing all instances of short tags being used in favor of more verbose alternative ``<? echo $variable ?>``


mAAdhaTTah:
@basteln3rk This is going to get closed. Sage by default requires PHP5.4+, where [short_open_tags are always available](http://www.php.net/manual/en/ini.core.php#ini.short-open-tag).

swalkinshaw:
What @mAAdhaTTah said.

austinpray:
Thank you for taking the time to put together a pull request. However, like @mAAdhaTTah said, Sage requires PHP 5.4+. 5.4 came out in 2012 and has even reached the end of its active support cycle.  http://php.net/supported-versions.php.

PR #1510, opened 2015-07-06
Added additional viewport settings, user-scalable, ua-compatibility.

kivlov:
For the sake of convenience, and something we over look on mobile devices.

### DIFF
--- a/lib/assets.php
+++ b/lib/assets.php
@@ -75,4 +75,130 @@ function assets() {
   wp_enqueue_script('modernizr', asset_path('scripts/modernizr.js'), [], null, true);
   wp_enqueue_script('sage_js', asset_path('scripts/main.js'), ['jquery'], null, true);
 }
+
+/**
+* Chrome on Android Meta Theme
+* 0.1
+* Author: Brave
+* Version: 0.1
+* Author URI: http://www.hellobrave.com
+**/
+
+function wp_head_chrome_android_meta_theme(){
+  $meta_theme_color = get_option('chrome_android_meta_theme',array());
+  $meta_theme_color['meta_theme_color'];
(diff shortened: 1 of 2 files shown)


JulienMelissas:
Ok first of all your second commit isn't something that belongs in Sage.

IMHO a site should NEVER include user scaling off by default. As for `chrome=1`, I'll have to do some research on that.

JulienMelissas:
Did a little bit of research on `chrome=1`... It looks like that is for support for chrome frame, which has been discontinued for a year and a half now. This is an answer from [this stackoverflow question](http://stackoverflow.com/questions/22059060/is-it-still-valid-to-use-ie-edge-chrome-1) from Feb '14:

> As for the chrome flag, you can leave it be if your users still use Chrome frame. As far as I can tell, despite support and updates for Chrome Frame ending, one can still install and use the final release: if you remove the flag, Chrome frame will not be activated if it is already installed. Once Google completely kills Chrome frame, the chrome=1 flag will do nothing more than consume a few bytes of bandwidth. Also, note that Chrome Frame is an open source project, so there's always a chance someone else might pick up were Google left off.

Because it's been discontinued, I'd vote no for adding support for that in, but thanks for the thought and contribution!

austinpray:
Yeah definitely do not turn off user scaling. That's an instant way to get your users to hate you.

swalkinshaw:
@kivlov this is a cool feature for some people but not something we're interesting in including in Sage by default. You could always create a repo/gist of it to share.

kivlov:
Thanks for the feedback fellas. The second pull request was accident, was wasn't aware of it. As for chrome=1, I only started using it this year. I found that without it 'theme-color' meta tags won't work. Will do some in-depth research on it.

PR #1516, opened 2015-07-17
*Add bootstrap dependency to manifest

pcfreak30:
Right now I don't believe bootstrap JS based components will work out-of-the-box as the files are not compiled together.

This fixes that.

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -4,6 +4,9 @@
       "files": [
         "scripts/main.js"
       ],
+      "bower": [
+        "bootstrap-sass-official"
+      ],
       "main": true
     },
     "main.css": {


Foxaii:
We use @austinpray's [asset builder](https://github.com/austinpray/asset-builder) so this is all handled automatically using bower's `main` property.

pcfreak30:
How does the main property auto include bootstrap? Not following that. I understood that it would pull all files defined in a components main, but its not defined in the manifest?

austinpray:
@pcfreak30 check out the "advanced usage with gulp" header over at https://github.com/austinpray/asset-builder/blob/29b60b1785cc166929380386573efa2d1f05e746/help/examples.md#advanced-usage-with-gulp

If there is anything I can explain better please give me feedback over on the asset-builder repo

PR #1427, opened 2015-04-14
Remove nav walker and Bootstrap navbar

retlehs:
Cleaner walker moved to a Soil module: https://github.com/roots/soil/pull/73 (**it no longer includes support for Bootstrap dropdowns**)

Remove the Bootstrap navbar and use a more generic header template (remember, this is a starter theme (and we're becoming framework agnostic))

image

PR #1517, opened 2015-07-17
*Refactor menu to better conform to bootstrap markup

pcfreak30:
There appears to be a few CSS classes missing that causes the menu to be a list, rather than a menu. It also isnt using a navmenu-header or brand class/div.

This fixes all issues in the menu.

It also should be considered to use `container-fluid` over `container`


Foxaii:
Please see #1427.

PR #1522, opened 2015-07-28
Update jscs validateJSDoc to jsDoc rule

chrisk2020:
Replaced depricated validateJSDoc rule with jsDoc rule: http://jscs.info/rule/jsDoc.html


austinpray:
Yep! Good to go.

PR #1523, opened 2015-07-30
Add `--fast` flag to gulp

QWp6t:
Allows users to skip minification by using `--fast` flag.

* `gulp --fast`
* `gulp watch --fast`

PS-
I'm not very good at naming branches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### HEAD
 * Add search templates ([#1459](https://github.com/roots/sage/issues/1459))
 * Allow `debugger` statements in development JavaScript ([#1487](https://github.com/roots/sage/issues/1487))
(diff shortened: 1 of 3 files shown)


austinpray:
Hmm if I don't get off my butt and do incremental builds we might have to merge this. My only hesitation is I don't want to add a flag and then have to deprecate it once the proper solution is in place. Gimme like 1 month at max and I'll refactor the gulpfile to support incremental builds and more.

QWp6t:
This dramatically improves build times for me on Windows. It takes 30 seconds or more for me to build. With this it usually takes less than a second, but sometimes as much as 2-3 seconds.

The changes you're talking about implementing will create a caching system for various parts of the build process, yeah? So we should hopefully see similar results in terms of reducing build times on Windows, yeah?

austinpray:
Incremental builds are coming very soon and will solve this. Don't want to introduce a flag and then deprecrate it immediately.

PR #1524, opened 2015-08-03
fix timezone offset in ISO 8601 date in byline

kevinlisota:
get_the_time('c') returns the local time in ISO 8601 format, but does
not fetch the related GMT timezone offset. This actually renders the incorrect time, as the timezone offset is always +00:00.

Google News, in particular, uses this time to show when an article was published. In our case it was always 7 hours too early because we are in GMT -7

get_post_time('c', true) is the correct function here, using the true parameter for GMT.


retlehs:
awesome, thank you! the WP.org codex page for `get_the_time` is super confusing.

kevinlisota:
I actually think the WP implementation for get_the_time('c') is simply wrong. If you fetch the post's published time in a friendly format, obviously it should use the local timezone set in WP, which it does. If you tell it you want ISO 8601 or similar, obviously it needs to retrieve the correct timezone offset. This has bugged us for month's on Google News. Finally figured it out.

PR #1527, opened 2015-08-13
Add keepSpecialComments: 0 to minifyCss process

nezencreation:
This removes all comments from the resulting compiled SASS and concat/minified CSS.  Solves an issue where the /*! style comments in bootstrap break the minification process.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -110,6 +110,7 @@ var cssTasks = function(filename) {
     .pipe(minifyCss, {
       advanced: false,
       rebase: false
+      keepSpecialComments: 0
     })
     .pipe(function() {
       return gulpif(enabled.rev, rev());


gulpfile.js
```diff
@@ -110,6 +110,7 @@ var cssTasks = function(filename) {
     .pipe(minifyCss, {
       advanced: false,
       rebase: false
```
austinpray: That's not valid javascript 

You need a comma after `rebase: false`

gulpfile.js
```diff
@@ -110,6 +110,7 @@ var cssTasks = function(filename) {
...
+      keepSpecialComments: 0
```
austinpray: That's not valid javascript 

You need a comma after `rebase: false`

QWp6t:
`/*!` comments are deliberately kept in place because libraries use that to provide their licensing info. Oftentimes licenses will require you to leave some sort of notice intact with all distributions of their code. You don't own the code and you're not always free to remove those notices.

Don't get me wrong. It's not my intention to arrogate to myself the position of arbiter for all of the free libraries out there on the internet. I'm simply clarifying why it would be irresponsible for us to do this by default in Sage.

PR #1541, opened 2015-09-15
Remove Modernizr

retlehs:


retlehs:
modernizr [finally came out with a new version](https://twitter.com/Modernizr/status/643464140397285376) (years later) and instead of updating to the new ver, i think it's probably better if we just remove it.

we've discussed this internally a few times recently. see this comment from @QWp6t:

> im in favor of getting rid of modernizr. i personally don't use it at all anymore. if there's anything i need to test for, i just write a small test for it. just look at the code for modernizr tests. most are self-contained. don't need the whole library for it. i guess if i was testing for a lot of stuff, i'd probably load it for convenience.

we're also not supporting less than ie9 out of the box, so the html5shiv isn't necessary.

JulienMelissas:
More often than not, I remove this at the end or beginning of the project. Totally in favor of removing Modernizr - it's obviously not very hard to add.

AustinGil:
I agree with removing Modernizr, but Im curious how you guys test for Javascript being enabled? I noticed base.php no longer has class="no-js" in it.

QWp6t:
By default, we don't. If you need it, just add the class back and then write some js to remove it.

If you want further support, please post on discourse.roots.io

AustinGil:
Gotcha. Well that's not too bad. Just wanted to make sure I wasnt missing something. Thanks for the reply and Ill make sure to post there in the future.

PR #1456, opened 2015-05-05
Allow revved production files on dev environment

colinduwe:
This change allows you to run `gulp --production` on your development environment and have the page load properly. `gulp --production` on dev breaks neat features of `gulp watch` like browsersync and so forth. However it does permit you to test the `--production` output locally before deployment.

PR #1554, opened 2015-10-10
Remove references to WP_ENV

retlehs:
WP_ENV is no longer required. previous dependencies on WP_ENV were removed in #1456 and #1382.


emzo:
Does this impact the [Stage Switcher](https://github.com/roots/wp-stage-switcher) plugin?

retlehs:
nope. that plugin relies on bedrock along with a custom `ENVIRONMENTS` constant, completely separate from the theme

emzo:
D'oh! Of course, thanks for the clarification.

PR #1555, opened 2015-10-10
Move init.php to config.php

retlehs:
this makes more sense.

also, we're down to 5 files in the `lib/` folder! :tada:


JulienMelissas:
Killer. People always ask me where they should put things and this should help reduce some of that confusion :+1:

PR #1557, opened 2015-10-10
Removing redundancy with WAI-ARIA in HTML

asuh:
Removed redundant ARIA roles
Added `nav-primary` class to main navigation


retlehs:
thanks!

ref https://discourse.roots.io/t/removing-redundancy-with-wai-aria-in-html/4941

Issue #1552, opened 2015-10-09
Reduce compile time of CSS bundles


aumo:
Currently, the whole styles are compiled twice:
- firstly as the `main.css` bundle
- secondly as the `editor-style.css` bundle because `main.scss` is imported in `editor-style.scss`.

Is there a justification or would it be enough to just concatenate, or include, the two files?

JulienMelissas:
@aumo - thanks for bringing this up. I looked into it a little bit yesterday and hope to have a better solution soon.

I totally agree that it seems to be pretty redundant, but I personally like the fact that I can add editor-only overrides after my site styles, so I'm not sure I want to get rid of that functionality.

Anyways, know I'm looking into it more and will report back.

JulienMelissas:
@aumo leave some thoughts here if you would please: https://github.com/roots/sage/pull/1560!

retlehs:
closed by #1560

PR #1560, opened 2015-10-13
Simplify/speed up editor style process.

JulienMelissas:
This is in response to https://github.com/roots/sage/issues/1552.

After playing around, I think the best solution is just to add a `_tinymce.scss` file in `/assets/layouts/` where we target the tinymce editor and can put conditional styles in there, and then just reference the main.css file in the wp editor.

This cuts styles compiling time of the default repo in half (roughly) because we're no longer creating 2 files, and no longer referencing a large scss file and injecting it into another one.

### DIFF
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -12,11 +12,6 @@
       ],
       "main": true
     },
-    "editor-style.css": {
-      "files": [
-        "styles/editor-style.scss"
-      ]
-    },
     "jquery.js": {
       "bower": ["jquery"]
     }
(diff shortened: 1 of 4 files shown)


chrisatomix:
Just backported this onto one of our in-build Sage sites to see if it works, the Styles compile time immediately dropped from **3.8s** to **2.3s** (very happy with that, it's a massive site). WP Editor is still working perfectly too.

JulienMelissas:
@chrisatomix great to hear it!!! Thanks for the feedback.

aumo:
That is what I ended up doing when looking for a way to reduce the assets compile time.

austinpray:
Looks good here

PR #1565, opened 2015-10-16
add auto height to .alignnone to prevent vertical stretch of image

88mpg:
I came across an issue with images being stretched in posts when using the class `.alignnone`. If the image's width is greater than that of the container's, the image will shrink to fit the container but the original height is still being declared in the img tag with `height=""`. As a result, the image is being stretched vertically. Adding `height: auto;` to `.alignnone` will resize the image vertically to maintain it's original aspect ratio.


retlehs:
thanks!

PR #1566, opened 2015-10-20
Remove ARIA text

jmslbam:
Because of https://github.com/roots/sage/pull/1557


retlehs:
there's still `<div class="wrap container" role="document">` & i think there's also other places where roles exist due to us declaring HTML5 theme support 

the [original thread](https://discourse.roots.io/t/removing-redundancy-with-wai-aria-in-html/4941) mentioned that the ARIA roles are already included in most of the elements we use by default, so technically ARIA roles are still a feature

PR #1569, opened 2015-10-28
Declare Text Domain in style.css

rob006:
Text Domain must be declared, if we want to translate custom page templates names.


retlehs:
thanks!

PR #1570, opened 2015-10-30
wp-classes alignment (right, left, center) height set to auto for proper scaling

88mpg:
Extending on https://github.com/roots/sage/pull/1565, `.alignleft`, `.alignright` and `.aligncenter` also skew vertically when the image width is wider than the container. Adding `height: auto` to these classes will override the `img` tag's height property and properly scale the image across different breakpoints.


retlehs:
thank you!

PR #1572, opened 2015-11-05
Add gulp-stylus, add to gulpfile

nathanhleung:
Stylus has greatly increased in popularity over the past couple of years—in 2012, it had about [6% market share](https://css-tricks.com/poll-results-popularity-of-css-preprocessors/) in terms of CSS preprocessing, Sass had 41% market share, and LESS had 51% market share.

Assuming repo "stars" are indicative of usage, [Stylus](https://github.com/stylus/stylus/) now has about 24% market share, [Sass](https://github.com/sass/sass) has 26% market share, and [LESS](https://github.com/less/less.js) has 50% market share.

It seems to me that if Sass is supported in Sage, we should also support Stylus.  Thoughts?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,6 +18,7 @@ var rev          = require('gulp-rev');
 var runSequence  = require('run-sequence');
 var sass         = require('gulp-sass');
 var sourcemaps   = require('gulp-sourcemaps');
+var stylus       = require('gulp-stylus');
 var uglify       = require('gulp-uglify');
 
 // See https://github.com/austinpray/asset-builder
@@ -95,6 +96,9 @@ var cssTasks = function(filename) {
         includePaths: ['.'],
         errLogToConsole: !enabled.failStyleTask
       }));
+    })
+    .pipe(function() {
+      return gulpif('*.styl', stylus());
     })
     .pipe(concat, filename)
     .pipe(autoprefixer, {
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "gulp-rev": "^6.0.0",
     "gulp-sass": "^2.0.1",
     "gulp-sourcemaps": "^1.5.2",
+    "gulp-stylus": "^2.1.0",
     "gulp-uglify": "^1.2.0",
     "imagemin-pngcrush": "^4.1.0",
     "jshint-stylish": "^2.0.1",
@@ -54,8 +55,5 @@
     "run-sequence": "^1.1.2",
     "traverse": "^0.6.6",
     "wiredep": "^2.2.2"
-  },
-  "dependencies": {
-    "gulp-stylus": "^2.1.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -54,5 +54,8 @@
     "run-sequence": "^1.1.2",
     "traverse": "^0.6.6",
     "wiredep": "^2.2.2"
+  },
+  "dependencies": {
+    "gulp-stylus": "^2.1.0"
   }
 }


QWp6t:
:+1:

I don't know about the others fella in Roots, but I'm cool with this. 

But `gulp-stylus` should be a dev dependency.

JulienMelissas:
I'm not sure I agree with the logic being this PR (I just don't think it's safe to make an assumption based off of repo stars, and you also forgot about adding libsass' stars to the Sass total), but in general, I think this is a good idea. The reason why there is support for Sass and Less is because someone could pull in a package with Bower using either or, and be able to use it no matter what preprocessor it uses. I don't see a reason to exclude Stylus, even though I don't personally use it.

Change to dev dependency (you can do that with `--save-dev-`) and I'm all for it :ok_hand:

JulienMelissas:
On the other hand, we are trying to un-bloat Sage, and I'm not sure another preprocessor is needed considering I haven't seen any questions on our discourse or anything like that...

kalenjohnson:
I've also never pulled in a Bower package with Stylus as it's preprocess or choice

nathanhleung:
The use case I was thinking was of sort of like on this [discourse](https://discourse.roots.io/t/how-to-correctly-add-a-custom-less-file-in-sage/3112/2)—adding custom styles using a specific preprocessor.  Also, I moved the package to devDependencies.

JulienMelissas:
Ok, just cause I'm a nerd, and didn't want to get real work done, I did some research:
I don't know if this is _better_ research than GitHub stars, but here's what I did:

I searched for 2 search terms, across GitHub repos "color" and "font", both things you would assume would appear to be commonly used terms across repos. Then I took the amount of code examples:

##"Font":
```
.sass/.scss = 5,179,808   [search](https://github.com/search?utf8=%E2%9C%93&q=font+extension%3Ascss+extension%3Asass&type=Code&ref=searchresults)
.less       = 4,814,475   [search](https://github.com/search?utf8=%E2%9C%93&q=font+extension%3Aless&type=Code&ref=searchresults)
.styl       = 163,790     [search](https://github.com/search?utf8=%E2%9C%93&q=font+extension%3Astyl&type=Code&ref=searchresults)
```

##"Color":
```
.sass/.scss = 6,207,814
.less       = 5,913,832
.styl       = 180,996
```

It seems that Stylus use is looking pretty bleak... Not sure this helps anyone's decision.

nathanhleung:
I guess Stylus isn't as popular as I thought...?  Or maybe people use it more for personal, unpublished projects?  I don't really know, but the decision's all up to you guys either way.

ciromattia:
For the same reasons shouldn't [PostCSS](https://github.com/postcss/postcss) be included as well? It's seems the future of CSS building, after all.
I mean, I'd stick with bare Sass support (even dropping Less, when BS4 will be released Less will lose a huge amount of users - IMHO of course) and maybe empowering the documentation with a bunch of recipes to activate one's preferred technologies, being them Less, Stylus, PostCSS, SVG icons and stuff like that.
Just my two cents ;)

QWp6t:
I see no problem with adding PostCSS as well. A couple guys in Roots already use it, so it would make sense to add it. There's not really a downside that I know of to adding these.

retlehs:
do we really want to add a bunch more dependencies?

ciromattia:
@retlehs just what I was saying ;)

JulienMelissas:
> On the other hand, we are trying to un-bloat Sage
 Me too!

retlehs:
closing for now. would like to revisit postcss in the future.

PR #1576, opened 2015-11-17
Stop pipe breaking on JS uglify error

vonkanehoffen:
Useful to not have the pipe break because Uglify chokes on your JS errors when you're using `gulp watch`.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -127,6 +127,7 @@ var cssTasks = function(filename) {
 // ```
 var jsTasks = function(filename) {
   return lazypipe()
+    .pipe(plumber)
     .pipe(function() {
       return gulpif(enabled.maps, sourcemaps.init());
     })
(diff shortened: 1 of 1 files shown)


retlehs:
this also came up recently: https://discourse.roots.io/t/js-build-error-gulp-work-gulp-production-fail/5268/2?u=benword

we should probably update the scripts task in the gulpfile to be like the styles one (which uses plumber):

```
// By default this task will only log a warning if a precompiler error is
// raised. If the `--production` flag is set: this task will fail outright.
```

/cc @austinpray

austinpray:
So this should definitely only be plumbed during watch. It should fail under all other circumstances. Having a fail silently could be deadly for automated deploys and such.

vonkanehoffen:
For me, the problem with with a fail in production was the site being left with no assets at all. `clean` gets run first before any failure happens (on the `default` task) so any dodgy JS being mistakenly deployed = a completely broken site. Not good!

Wonder if there's an easy way to only run `clean` after everything else has finished successfully?

mAAdhaTTah:
The deployment process should abort or roll back if it encounters an error during the build process. That's where your issue really is.

austinpray:
@vonkanehoffen You could also avoid this by running gulp task in a "before" hook in whatever deployment solution you have. It should make sure the assets build before even starting to deploy the latest changes.

But yeah also what @mAAdhaTTah said.

retlehs:
focus is on sage 9 right now

Issue #1429, opened 2015-04-17
Templates folder


schrapel:
This folder structure is way better than having a bunch of files in the root with markdown, bower, etc. What do you think?

```
templates/
│   404.php
│   page-home.php   
└─── partials/
    │   header.php
    │   footer.php
└─── layouts/
    │   base.php
    │   base-home.php
```

schrapel:
For reference, I learnt about it here: https://github.com/dxw/govsite/

schrapel:
The only issue I see, is that when you move functions.php it won't be able to locate_template the /lib files since WP updates TEMPLATEPATH value to theme/templates folder.

QWp6t:
For the record, i love the restructuring ideas implemented in govsite. Kudos @dxw.

schrapel:
@QWp6t I would happily implement and submit a pull request myself but I think we need to discuss what to do with functions.php. We have three options I can see

1. Replace locate_template with a function to check if the file exists and move to templates folder
2. Keep functions.php outside template folder (although I'd rather keep all PHP together)
3. Move functions.php to lib folder and rename to loader.php, and have a functions.php inside the templates folder that loads the single loader file.

retlehs:
would love to do something like this. it's time to start [publicly] talking about next steps regarding PHP in the theme.

if we are going to be adding functionality like this, it would ideally be managed outside of the theme itself. the best solution is probably to add composer to sage. that unfortunately would bring our package manager count to 3 - npm, bower, and composer.

by adding composer to sage we could also separate the following components into their own repos/projects:

* the theme wrapper
* conditional tag check

related reading: http://tomjn.com/2014/04/01/theme-composition-vs-frameworks/

we've discussed this quite a bit internally but haven't made any decisions yet.

retlehs:
playing with this right now...

wordpress requires functions.php, style.css, and screenshot.png to belong in the subdirectory with the primary wordpress templates (index.php, etc). it would be odd to have those files in a dir called `templates/` when they're not actually templates

schrapel:
Yeah I know. That is why I didn't actually do anything or submit a pull request.

retlehs:
seems like we could modify the theme wrapper and also use a utility like #1414 to replace `get_template_part` to get this working without the weird theme file requirements

like i mentioned in my previous comment, this is going to require us adopting composer in the theme and separating out components.

JasonTheAdams:
[...]

PR #1590, opened 2015-12-19
[WIP] Sage 9

retlehs:
work in progress, no ETA for release at this time

## goals

* [x] restructure theme, use autoloader (@QWp6t)
* [x] new templates folder organization — #1429 (@QWp6t)
* [x] switch to eslint (@QWp6t)
* [x] psr-2 coding standards — #1594 (@QWp6t)
* [x] new front-end workflow — introduce webpack, drop gulp and bower in favor of npm (@ptrckvzn)
* [x] bootstrap 4
* [ ] add Laravel Blade support
* [ ] fix autoloader/composer stuff
* [ ] docs updates


QWp6t:
`npm WARN deprecated gulp-minify-css@1.2.3: Please use gulp-cssnano instead.`

joaovpmamede:
I guess even gulp could be dropped in favor of something like:

[browser-sync-webpack-plugin](https://www.npmjs.com/package/browser-sync-webpack-plugin)
or by looking at one recipe from browserSync repo (e.g. [webpack.monkey-hot-loader](https://github.com/BrowserSync/recipes/tree/master/recipes/webpack.monkey-hot-loader))

rob006:
> drop bower in favor of npm

Why?

austinpray:
@rob006 with webpack you can use npm, bower, and or manually vendored scripts together with no problems. However, going forward, we are going to recommend using npm. The packages in npm's registry are higher-quality and usually built with UMD/module loading in mind.

As far as community sentiment: a lot of the react community doesn't even bother to publish to bower. [Angular 2 dropped support for bower as well](https://github.com/angular/angular/issues/4018).

rob006:
@austinpray I'm just curious what is wrong with bower so people want use package manager designed for JS for all frontend dependencies (which may not contain any JS content, for example css frameworks or font sets). And I always get some irrational arguments like "bower's tree is flat" or "bower is unreliable because depend on GitHub, and GitHub can be DDoSed". So Sage drop bower because some specific issue, or just because npm is cooler? 
Personally I pretty like separation of PHP dependencies (composer), frontend dependencies (bower) and node dependencies which are used only as development tools (npm).

> Angular 2 dropped support for bower as well.

No, it doesn't, it just not support all features of bower. It even use bower:
https://github.com/angular/angular/blob/master/bower.json
https://github.com/angular/angular/blob/master/package.json#L46

evanfuture:
There have been a lot more times recently when a package I really wanted to use just wasn't available on bower, and I had to add it manually to my overrides files instead of just installing it.  Its not that you can't keep using bower, but I'd say that most things you'd install that have a bower.json file will also have a package.json file.

QWp6t:
@rob006, I think you might have missed apray's opening sentence. Here it is again.

> **with webpack you can use** npm, **bower**, and or manually vendored scripts together **with no problems**

So feel free to keep using Bower if that's what you're into. But we will most likely be pulling Sage 9 dependencies from npm.

rob006:
@QWp6t I have not missed that, I just ask what is the reason of dropping bower and what is npm advantage. I'm not a bower fanboy, if there is something wrong with bower I'd like to know that. If it's only personal preferences, just say it, I'm fine with it.

austinpray:
@rob006 a 100% logical reason to drop bower is that it is not needed. For the purposes of sage as it appears out of the box: npm and bower have overlapping features. We simply do not require 3 package managers.

If you personally enjoy bower or your project calls for bower, it will still be just as easy to use. If not easier than before.

retlehs:
@ptrckvzn has been working on webpack with sage 9 :heart_eyes:  more info: https://discourse.roots.io/t/sage-9-webpack/6066

package.json
```diff
@@ -18,41 +18,52 @@
     }
   ],
   "scripts": {
-    "build": "bower install && gulp",
-    "jshint": "gulp jshint",
-    "jscs": "jscs gulpfile.js assets/scripts/*.js"
+    "build:production": "export SAGE_ENV=production && webpack",
+    "build": "export SAGE_ENV=development && webpack",
+    "watch": "export SAGE_ENV=development && node dev.js"
```
QWp6t: I'm pretty sure none of these will work in Windows. Users will just get an error saying that it can't find the command `export`, or worse they might have an export command, but it does something completely different and yields wildly unexpected results.

swalkinshaw: The exports can be removed. They aren't needed when defining an env var
inline like that.

QWp6t: That still wouldn't work on Windows. Then it would just say it can't find the command `SAGE_ENV=development`. Windows uses the `set` command to set environment variables, i.e., `set SAGE_ENV=development`

ptrckvzn: I made two more [commit](https://github.com/roots/sage/compare/sage-9...ptrckvzn:webpack-sage-9?expand=1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R21) on my branch. I removed the variables for two of the scripts. I can  work to remove it from the watch script too

henkler:
There appears to be an issue with autoloader and namespacing when including Sage 9 in as a composer dependency for a larger project.

I've got a stock clone of the Sage 9 branch that I'm then pulling in as a composer module into a bedrock project (via the main composer.json).  After doing a "composer update", WordPress throws errors about not being able to find App\add_action() and other WordPress global functions.  Could the generated autoloader code be forcing the App namespace and not defaulting to a recursive search to the global namespace?

If Sage 9 is just dumped into the /web/app/themes directory (followed by a composer update in the theme), it works just fine.

I believe Sage 8 works fine when being installed as a composer dependency.

screen shot 2016-03-17 at 1 07 53 pm

alan-c:
Possibly consider deleting `.babelrc` and putting the preset into `webpack.config.js`, as in: 

```
query: {
  "presets": [ "es2015"]
}
```

Gets rid of an otherwise unnecessary file.

henkler:
Fixed the issue with autoloader when pulling in Sage into Bedrock via composer.json.  See #1628

schrapel:
We are loading $sage_includes before Composer but $sage_includes requires Roots\Sage\ classes that are autoloaded

schrapel:
I have custom classes on a new namespace loaded with PSR-4 through composer. However, they only work when I load composer before $sage_includes where they are being called. How is it possible the Roots\\Sage\\ namespace loads earlier and my namespace doesn't?

aried3r:
> Possibly consider deleting .babelrc and putting the preset into webpack.config.js, as in:

Don't all kinds of other tools depend on it, like mocha? I've seen a few projects where they keep the `.babelrc` because all of their tooling accesses it.

alan-c:
> Don't all kinds of other tools depend on it, like mocha? I've seen a few projects where they keep the .babelrc because all of their tooling accesses it.

I've seen arguments like that for keeping `.babelrc` in projects. But it just felt like a build process configuration should be kept with the rest of the build process configurations.

But whether it is decided to put the Babel configs in `webpack.config.js` and require overriding via a `.babelrc` for external tools _or_ put them in a `.babelrc` and require overriding inside `webpack.config.js` for build process configurations... half a dozen of one, six of the other.

Issue #1361, opened 2015-03-02
Travis CI: Test with NodeJS v0.12.x


ntwb:
If your going to require NodeJS 0.12.x c5587b6c3415ec70e08a9741f0a3d5832c4c0668, Travis CI  should test using NodeJS 0.12.x, currently Travis CI is testing against the latest v0.10.x release.

https://travis-ci.org/roots/sage/jobs/52636851#L29
```
Pre-installed Node.js versions
v0.10.36
```

I don't think this is possible at the moment as NodeJS v0.12.x is only available "on demand" for NodeJS projects, NodeJS is pre-installed for all languages but only the "current stable release".

See http://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Provided-Node.js-Versions

It looks like April will be when non NodeJS projects will start using Node JS v0.12.x
https://github.com/travis-ci/travis-cookbooks/pull/452#issuecomment-74512202

swalkinshaw:
Good point. Requiring 0.12.x was mostly to ensure the latest npm version. We're more concerned about npm than Node.js and figured it's easier to get them bundled.

We'll keep this open as a reminder to update Travis.

ntwb:
Scott, to be pedantic and hopefully not annoying ;)

The version of NPM used by Sage is determined by `npm install -g npm@latest` via 8c374e2136294d3357ecc8a13d6563e5341e09bd

In fact NodeJS v0.12.0 only ships with NPM v2.5.1, the latest NPM is v2.6.1

So nothing to see here, once Travis CI has updated Node JS for non-NodeJS projects check what version is used by default, if it is v0.12.x then close this ticket, if not create a pull request. :)

swalkinshaw:
No problem. I didn't literally mean `npm@latest` by "latest". I just meant we want to people to be on 2.x.

ntwb:
:+1:

intelligence:
There seems to be some issues with 0.12 
https://github.com/npm/npm/issues/7349

thinkstylestudio:
Regarding 

> "No problem. I didn't literally mean npm@latest by "latest". I just meant we want people to be on 2.x." 

In the tutorial on the website states: 

```npm install -g npm@latest```

I took that as verbatim and i'm now having errors (I'm wondering because the above command actually does something if others are not getting the same type of errors?... that's if the errors Im' having are even connected to this discussion):

```
annot read property 'dependencies' of undefined
npm ERR!
[...]

PR #1591, opened 2015-12-20
Update dependencies, switch to eslint

QWp6t:
Note: I'm targeting sage-9 branch. This isn't going to master. J

Using the latest versions of npm and node.js, I was able to update all of the npm dependencies to their latest stable versions. I then tested by running gulp. Everything passes, as I'm sure travis will confirm.

I switched to eslint because I'm pretty sure we all agreed that it was the right move, but also because jshint has a history of playing fast and loose with semver, which makes it difficult to have it as a dependency. Since eslint can do everything jshint can do (except validate json files, apparently -- note that bower.json is no longer validated in `gulp lint` task) and it also does code style, I figured it was a better choice.

I notice that travis failed. I'm not sure how to get travis on a newer version of npm (or node.js?), but you can test this branch yourself; everything should pass.


.travis.yml
```diff
@@ -1,24 +1,37 @@
...
+    - g++-4.8
 
 cache:
   directories:
     - bower_components
     - node_modules
+    - vendor
 
 install:
+  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
```
swalkinshaw: Makes a little more logical sense to put `rm -rf ~/.nvm` in `before_install`.

swalkinshaw: is this needed because we're using PHP as the main language? Because Travis does support nvm natively.

swalkinshaw: `nvm` does already exist though if that makes it simpler: http://entulho.fiatjaf.alhur.es/guias/how-to-use-node-along-with-other-language-on-travis-ci/

QWp6t: I'm assuming it's needed. I'm not sure, though; I didn't test without it. I did see that Travis supports nvm natively, but I assume that's only for node_js containers. 

I poached this from some random nerd online: http://austinpray.com/ops/2015/09/20/change-travis-node-version.html

swalkinshaw:
:+1: for eslint

ntwb:
Agree with eslint also :+1: 

Noting this PR will also close https://github.com/roots/sage/issues/1361

austinpray:
Eslint 👌🏻

Did you make sure the jscs rules don't conflict with the eslint recommended rules? Do we need to be using both?

PR #1592, opened 2015-12-21
Add PHP 7.x "nightlies" to Travis CI and allow to fail

ntwb:


ntwb:
Coinciding with my comment over here -> https://github.com/roots/sage/commit/e1c23a49320cad9ea592d6af821272be2185972d#commitcomment-15088666 on allowing PHP 5.5 Travis tests to fail I added "nightly" to do the same which at the time of writing tests against `7.1.0-dev`

QWp6t:
I'd be in favor of re-adding HHVM before PHP nightly, but I left out HHVM because I wasn't sure how concerned we should be with the build matrix. I'll defer to the rest of the team on this one.

ntwb:
Cool... It was as quick to create the PR as it would've been to just create an issue :smirk: 

On the HHVM front, the current Ubuntu 12.x Travis CI infrastructures latest version of HHVM available is `3.6.6`, since `3.7.x` Facebook stopped compiling HHVM for Ubuntu 12.x.

I'm working on setting up HHVM for WordPress.org with Travis CI that uses the new Travis CI infrastructure with Ubunutu 14.x and Docker so that the latest HHVM `3.11.x` can be added to the matrix. Once I've got this done I'll open up a PR here for the same for Sage.

PR #1594, opened 2015-12-28
PSR-2 and other PHPCS rules

QWp6t:
Note: Targeting sage-9 branch.


.travis.yml
```diff
@@ -41,5 +41,6 @@ install:
 
 script:
   - npm run build
-  - phpcs --standard=ruleset.xml --extensions=php --ignore=node_modules,bower_components,vendor -n -s .
+  - phpcs --report=summary
+  - phpcs --report=summary --standard=phpcs-templates.xml -n
```
QWp6t: Note: `-n` means that travis will ignore phpcs warnings, only errors will trigger failures.

This gives us a bit of discretion.

PR #1599, opened 2016-01-05
Added `stability` flag to example Composer install command

grega:
Composer's `stability` flag defaults to `stable` (see https://getcomposer.org/doc/03-cli.md#create-project) which causes the command `composer create-project roots/sage web/app/themes/your-theme-name-here` to fail with the error:
```
[InvalidArgumentException]
Could not find package roots/sage with stability stable.
```
Using `composer create-project roots/sage web/app/themes/your-theme-name-here --stability dev` resolves this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ If you're already [using Composer to manage WordPress](https://roots.io/using-co
 The example below assumes you're using Bedrock. If you're not, simply change the target path accordingly.
 
 ```sh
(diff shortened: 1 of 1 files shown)


theblackhole:
:+1: same issue, adding `--stability dev`  works

swalkinshaw:
This is a problem with Sage on [Packagist](https://packagist.org/packages/roots/sage) right now. Once we fix it to properly show the tags/versions, it will default to the latest one.

PR #1602, opened 2016-01-19
gulp-minify-css (deprecated) -> gulp-cssnano

builtbylane:
the exact warning: 
`npm WARN deprecated gulp-minify-css@1.2.3: Please use gulp-cssnano instead.`


austinpray:
Are those options that are being passed to CSS nano still relevant?

builtbylane:
Great question @austinpray! Nope, and even better news... autoprefixer can be run as a transform right in cssNano, so we can actually drop gulp-autoprefixer. to keep things clean, I'm going to close this and resubmit.

PR #1603, opened 2016-01-21
Disable JS variable renaming when sourcemaps are enabled

jmarceli:
Disable variable renaming when JS sourcemaps are enabled in order to ease JS debugging with browser debugger (e.g. Chrome Console).


austinpray:
We can't do this because if we aren't mangling in dev we should just disable uglify in dev entirely. I don't want to disable uglify in dev right now. As such: the underlying issue is the sourcemaps that need to get fixed.

Issue #1609, opened 2016-01-31
Use gulp-cssnano instead of gulp-minify-css


JakeCobley:
gulp-minify-css was deprecated at version 1.2.3

JakeCobley:
#1610

landerss0n:
+1

PR #1610, opened 2016-01-31
Removed deprecated gulp-minify-css in favour of gulp-cssnano

JakeCobley:
gulp-minify-css was deprecated at version 1.2.3

Issue: #1609


gulpfile.js
```diff
@@ -104,7 +104,7 @@ var cssTasks = function(filename) {
         'opera 12'
       ]
     })
-    .pipe(minifyCss, {
+    .pipe(cssNano, {
       advanced: false,
```
austinpray: Are these options still relevant?

JakeCobley: As far as I'm aware advanced: false can be removed. That was an oversight on my behalf.

austinpray: While you are revising the options: basically disable all "clever" options. Potentially destructive optimizations are not wanted.

JakeCobley:
Updated commit to remove unneeded options.

austinpray:
Looks good.

@landerss0n does this PR work for you on your projects as well?

landerss0n:
Yes it works fine

austinpray:
>While you are revising the options: basically disable all "clever" options. Potentially destructive optimizations are not wanted.

JakeCobley:
Added `safe: true` option, should be sound now.

PR #1614, opened 2016-02-04
Remove useless line breaks

diego-betto:


retlehs:
thanks. can you re-submit this PR for the sage 9 branch? not really making any more changes to version 8.

diego-betto:
Seems base.php it's ok in sage-9 branch. Thank you.

PR #1618, opened 2016-02-14
gitignore composer vendor directory

lukasbesch:
`composer create-project` (see 6a6ff885360a96b70d9e883f24ab73435762dbc4) adds vendor directory which we do not want to have in version control.

composer.json and composer.lock should probably also be ignored – at the moment they are only covered by Bedrock's gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist
 bower_components
 node_modules
 npm-debug.log
+vendor


retlehs:
thanks

> composer.json and composer.lock should probably also be ignored

these files will be relevant in sage 9

PR #1628, opened 2016-03-25
Fix autoloader load order issue when including Sage 9 via composer.json in Bedrock

henkler:
The Composer Autoloader has issues with the "files" directive in composer.json when Sage 9 is included as a package in a Bedrock project via the root composer.json.

After diagnosing, I don't see a way the "files" directive can be used in the Sage composer.json and not cause issues with load order.  The root composer package (in Bedrock) will see the files directive in the package and automatically attempt to load prior to WordPress. 

psr-4 autoloading for Roots\Sage should still work fine in this scenario since these are lazy-loaded.

I've moved the offending 4 files from autoloader back into functions.php the way they were in Sage 8.  I'm not sure if there is a more clever way to do this, but it fixes the issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,13 +26,7 @@
   "autoload": {
     "psr-4": {
       "Roots\\Sage\\": "src/lib/Sage/"
-    },
-    "files": [
-      "src/helpers.php",
-      "src/setup.php",
-      "src/filters.php",
-      "src/admin.php"
-    ]
+    }
   },
   "require": {
     "php": ">=5.5.0",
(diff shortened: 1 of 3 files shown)


retlehs:
thanks!

aaemnnosttv:
@retlehs @henkler You also need to make the [require autoload](https://github.com/roots/sage/pull/1628/commits/a994889f9a0f72bd52a0b74f17713f890eabd07e#diff-78cd5aa3783a74555c9938a2a81d01c6L29) conditional to the existence of the file otherwise this will fatal since that file will not exist if the theme were installed as a composer dependency ;)

retlehs:
@aaemnnosttv thanks :+1: 508de7dd3bc31142c4e3689ac2a74077f997e4fa

PR #1630, opened 2016-03-30
Custom logo support

peiche:
Adds custom logo support for WordPress 4.5. http://codex.wordpress.org/Theme_Logo


retlehs:
no longer making changes/updates to sage 8. we can consider this for sage 9, but we wouldn't support WP < 4.5 for this feature so the `function_exists` check isn't necessary

i'm also unsure about supporting this out of the box considering wordpress still doesn't support SVG uploads

johnny-bit:
Hey @retlehs - supporting this out of box does no real harm. Most people are totally fine with having logo as png (or worse - jpg) image. Those that need full svg support, can and will install plugin for svg support (i'd simply add one as a dependency in composer.json).

retlehs:
> supporting this out of box does no real harm

agreed

> Most people are totally fine with having logo as png (or worse - jpg) image.

stats on this statement? i believe most sage users are doing logos as SVG from the theme files based on what we see from new sites being shared on discourse

> Those that need full svg support, can and will install plugin for svg support (i'd simply add one as a dependency in composer.json).

sure... but these users are most likely to not use the custom logo WP implementation and will continue to serve SVGs from the theme for logos. 

this makes a lot more sense for people making themes to sell or distribute, which isn't our primary demographic.

johnny-bit:
> stats on this statement? i believe most sage users are doing logos as SVG from the theme files based on what we see from new sites being shared on discourse

Most people as in people using themes, not developing them. I had my share of customers wanting to upload graphics, logos etc after buying theme. However, surely made-for-customer theme, with perfect branding etc, it makes more sense to blend logo right in theme.

> sure... but these users are most likely to not use the custom logo WP implementation and will continue to serve SVGs from the theme for logos.

You mean developers, and I fully agree. I do that too.

> this makes a lot more sense for people making themes to sell or distribute, which isn't our primary demographic.

Agreed. still - no harm in it simply _being there_, just to satisfy those wanting "all new Wordpress features". I mean, without `function_exists` wrapper this would be 2 lines additional in total (excluding possible fallbacks) and one could simply remove them if not needing them or add them if needing (if this or similar PR won't make it into sage 9).

Unrelated fun side-note: i'm arguing here about no-harm of feature i'd likely disable or wouldn't use for ~50% of themes i plan on doing once sage 9 is released... Or would use, but it won't be used by customer because fall-back will be "exactly what is needed". So... Only sake of this is "completeness" I guess?

QWp6t:
Closing this. 
* Not making changes to Sage 8. 
* Fails CI. 
* Echoing a function that returns void. 
* Lack of interest.

PR #1640, opened 2016-04-30
Feature/manage virtualhost with dotenv

luism-s:
**The problem:**
If I'm using Sage with a team of developers, all of us have to set the same virtualhost's name in each of our machines to match the `assets/manifest.json` file. 
If I, for example, set my virtualhost url to project-name.dev, and my team mate sets his to projectname.local and I pull the changes he made in  `assets/manifest.json`, my browsersync wont work because now I have the wrong virtualhost url.

**My solution:**
I propose the use of [dotenv](https://www.npmjs.com/package/dotenv) to specify the virtualhost url in each project. 
This way, only a file called .env.example will be passed in git, and then each team member copies that file to another named .env (which will be ignored by git) and set his own virtualhost url.
The usage of dotenv will of course also enable other local variables to can be used with the browsersync plugin, such as port and host options.

This solution has been tested, and I'm also currently using it in my own front-end boilerplate project.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+DEV_URL=virtualhost_url
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,8 @@ bower_components
 node_modules
 npm-debug.log
 vendor
+
+# Dotenv
+.env
+.env.*
+!.env.example
--- a/README.md
+++ b/README.md
@@ -101,23 +101,15 @@ You now have all the necessary dependencies to run the build process.
 
 ### Using BrowserSync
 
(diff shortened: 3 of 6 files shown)


swalkinshaw:
Our opinion on issues like this has always been that your team should be using the same host name. We know that not every team is set up that way to begin with, but we don't to want complicate the process for everyone else. This solution requires:

* a new dependency
* a new file
* an extra configuration step

If there was some middle ground or smaller change that would make this more flexible, we'd consider it. But not as this PR currently stands.

luism-s:
> some middle ground or smaller change that would make this more flexible

What if the gulp script first checks if there is a .env to read, and if there is not, then the devUrl in manifest.json would be used.

PR #1642, opened 2016-05-02
Sage 9 - fix bootstrap4 styles 

PTiCA1:


retlehs:
what is this fixing?

aaemnnosttv:
Class changes in the v4 alpha which break `@extends` for a few styles due to the parent class no longer existing.

- `.btn-default` is now `.btn-secondary`
- `.thumbnail` is now `.img-thumbnail` (I think?) but @PTiCA1 has something slightly different so I'm not sure if it's the same.

PTiCA1:
@aaemnnosttv Yes exactly.
The new version of the bootstrap class names are slightly different.

I just class .thumbnail replacing figures + alternatives .img-thumbnail
http://v4-alpha.getbootstrap.com/content/figures/

In the current version of sage 9 can not perform "run NPM build" with error @extend ...

assets/styles/components/_wp-classes.scss
```diff
@@ -32,10 +32,14 @@
 
 // Captions
 .wp-caption {
-  @extend .thumbnail;
+  @extend .figure-img;
+  @extend .img-fluid;
+
+  // alternative styles
+  // @extend .img-thumbnail;
```
retlehs: can you remove these two comments?

retlehs:
gotcha, thanks. remove those commented out lines and then this is good to merge

PTiCA1:
removed...

PR #1661, opened 2016-06-04
watch.js/browser-sync: use dev-ip instead of hard-coding localhost

ptrckvzn:
Hard-coding localhost prevented access to BrowserSync from other devices on the local network.


retlehs:
http being hardcoded is also going to be an issue with sites that use SSL on local dev

ptrckvzn:
True. It does not work with Trellis SSL set to self-signed... Tried hard-coding https with "self-signed" for testing but it did not work either. It's strange because Safari does not mind if the publicPath is relative or left to default - assets are loaded. Chrome gives an ERR_INSECURE_RESPONSE in the console for every assets loaded.

ptrckvzn:
The problem seems to be the css-loader with source maps enabled. If css in inlined (when running 'npm run watch'), It will not load css assets loaded with url(...) unless the [public path is absolute](https://github.com/webpack/css-loader#sourcemaps) (more info [here](https://github.com/webpack/css-loader/issues/29)).

The solution is to load all css assets with url-loader to inline them in the bundle or dynamically assign the public path. I'm trying the later here : https://github.com/ptrckvzn/sage/commit/d4256ae97978b9c36959f6dbd2891087699b49fa

This way, `watch.js`, `assets/config.json` are simplified and it works well with either http or https.

What do you think of this method?

retlehs:
@ptrckvzn lgtm :+1:

ptrckvzn:
I will close this and open a new PR.

Issue #1540, opened 2015-09-15
Gulp Notification on Error


christianmagill:
How about adding a notification outside of the console on error during watch using gulp-notify or similar?

retlehs:
not something we'd be looking to add to the theme by default. if you implement this, i'm sure others would find it useful if you shared over at https://discourse.roots.io/

PR #1664, opened 2016-06-10
add gulp-notify message on css or javascript error

flurinduerst:
In addition to the error message in the console, shows a notify-error. This is especially useful for those who do not have the console active at all times, as you will instantly be notified if something went wrong. 

image


QWp6t:
https://github.com/roots/sage/issues/1540

PR #1665, opened 2016-06-10
Remove unnecessary string concat

hulet:
String concat is not required here.

Reduce complexity! Improve speed! :) This also makes tools like PHPCS happy.

See da514346 for the origins of this double-string concat.

### DIFF
--- a/lib/assets.php
+++ b/lib/assets.php
@@ -46,7 +46,7 @@ function asset_path($filename) {
   static $manifest;
 
   if (empty($manifest)) {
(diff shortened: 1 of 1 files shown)


retlehs:
thanks, but we're no longer making changes to sage 8. this doesn't exist in sage 9:

https://github.com/roots/sage/blob/764d5fb686603f1ca94c94c44c4ef7da5ebb4d1a/src/lib/Sage/Asset.php#L36

PR #1670, opened 2016-06-22
Set private false

Metalchocobo:
It can be set  set private to false to register package in the bower registry?
Thanks!


retlehs:
thanks, but we're no longer using bower in [sage 9](https://github.com/roots/sage/tree/sage-9) and we're not really accepting PR's for sage 8 anymore

PR #1649, opened 2016-05-19
Added Bootstrap Navwalker for main menu

lyubb:
Added wp-bootstrap-navwalker http://twittem.github.io/wp-bootstrap-navwalker/ "A custom WordPress nav walker class to fully implement the Bootstrap 3.0+ navigation style in a custom theme using the WordPress built in menu manager."

Issue #1475, opened 2015-05-26
nav_walker?


lgt:
As I remember Roots has included a nav_walker. There is no more such a support  for bootstrap menus?

Foxaii:
See #1427 and [here](https://discourse.roots.io/t/how-to-restore-the-bootstrap-nav-walkthrough/3657/).

In future please keep questions like this to the Discourse forum only.

austinpray:
It's been removed: https://github.com/roots/sage/pull/1427

PR #1676, opened 2016-07-08
Added Bootstrap Nav Walker +  Navigation Header to the default theme.

mavieth:
I added a default bootstrap nav walker to the theme template. I have found that each time I use the Sage theme, I end up using the default Bootstrap navigation, and figured this would be nice for people who do not know how to implement it at first.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -10,12 +10,13 @@
  * @link https://github.com/roots/sage/pull/1042
  */
 $sage_includes = [
-  'lib/assets.php',    // Scripts and stylesheets
-  'lib/extras.php',    // Custom functions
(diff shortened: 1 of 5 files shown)


AdamWills:
This has previously been taken out of the project as there's a push to make Sage more framework agnostic.

ptibbetts:
It was removed in https://github.com/roots/sage/pull/1427 

Somebody asked about it in https://github.com/roots/sage/issues/1475 and someone else re-added it in https://github.com/roots/sage/pull/1649

There's nothing stopping you from keeping it in your own fork of Sage though 😃

PR #1679, opened 2016-07-16
Keep inline with the DRY principle.

allanchau:
- Moved page header to the base template.
- Added filter to hide page header.


QWp6t:
The page header is an isolated component that is intended to be reused. The header, footer, and sidebar, on the other hand, are all structural blocks, which is why we keep them DRY. The primary purpose of the page header is to display content, not structure the layout.

allanchau:
Fair enough :)

PR #1643, opened 2016-05-04
README updated for running browsersync without bedrock

Jeger:
Readme now has instructions on how to run `npm watch` without bedrock

PR #1681, opened 2016-07-17
Automatically determine publicPath (WIP)

QWp6t:
Related: #1643

I haven't tested this too much (read: I only tested it with Trellis + Bedrock + Sage default stack config). 

The way this works is it connects to the server and looks for any scripts or styles being called from a path with the same name as the theme directory (e.g., "themes/sage"), and then it does a bit of parsing and string manipulation and spits out what it thinks is the correct path.

If you're going to test this, don't forget to run `npm install` first since I added two new dependencies (`cheerio` and `sync-request`).


QWp6t:
I'm a bit of a dipshit when it comes to node/npm. There is most likely a cleaner way of doing this, and perhaps this should only be done when loading the watcher, not when building (so we don't have to have dev server running in order to build). And it might be possible to do this asynchronously, but I'm not sure. In short, it would be helpful if @austinpray or anyone else who knows more about this stuff than me could have a look at this and expand on it (or let us know if it's complete trash and we should be doing something else).

swalkinshaw:
I don't know about this feature in general, but this implementation is a non-starter. The config needs to remain in JSON format. Just because it's a subset of JS doesn't mean we can change it to that.

QWp6t:
This PR was made in response to Ben's comments here: https://github.com/roots/sage/pull/1643#issuecomment-216881614

The only way I can think of that you can reliably achieve the goal of never having to set the publicPath manually would be to sniff it out, which is what I'm doing here.

I'm fine with keeping it as something that users manually configure. They're already in there setting their domain name, so it's not a big deal to just have them configure the path too.

Also, all I did was brought the JSON into the JS. You could very easily just put the JSON back into its own file. It's not that big of a deal, i.e., line 6 could be changed to `var CONFIG = require('./config.json');`.

swalkinshaw:
Yeah, that's all fine. I'm just saying whatever the implementation, `config.json` should remain as is.

I think any automatic detection of this should be moved to `watch.js`. Then remove:

```json
  "devUrl": "http://example.dev",
  "devPort": 3000
```
from `assets/config.json` by default but let them be overridable from there.

PR #1683, opened 2016-07-22
Set devUrl from Bedrock's .env (if it exists)

leek:
If Bedrock's `.env` file exists, we read it and set `config.devUrl` to the `WP_HOME` value from it. Otherwise, we continue as normal (using the value from `manifest.json`).

This will allow a seamless integration of Bedrock and Sage without compromising the flexibility of using Sage without Bedrock.

This is similar to #1640 but much simpler and is focused solely on using Sage with Bedrock (a common scenario, I assume).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -31,6 +31,10 @@ var path = manifest.paths;
 // `config` - Store arbitrary configuration values here.
 var config = manifest.config || {};
 
+// Use the URL defined in Bedrock's .env (if it exists)
+require('dotenv').config({silent: true, path: '../../../../.env'});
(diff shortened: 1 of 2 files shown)


QWp6t:
We appreciate that you took the time to submit a PR, but unfortunately we can't accept this one.

1. We're not adding new features to Sage 8. Our focus is on Sage 9 at the moment.
2. Bedrock is switching over to INI. https://github.com/roots/bedrock/issues/240

You can try resubmitting your PR to target Sage 9 if you'd like (along with a plan to migrate to INI parsing), where it can be discussed whether to add this feature. For the record, I'm personally against this. I don't think we should be sniffing around the file system anywhere outside of the sage directory.

PR #1684, opened 2016-07-23
Fix path to stylesheet in add_editor_style

johnnyicarus:
String passed to asset_path needs to contain the directory name, otherwise it 404s on the post/page admin screen.


retlehs:
thanks!

PR #1685, opened 2016-08-04
Update Asset class to allow assets in subfolders

igniteonline:
properly analyse manifest and allow for assets to be stored in subfolders within dist (such as admin/style.css or similar).


QWp6t:
I went a slightly different direction with this, and it should no longer be an issue. Thanks for bringing it to our attention. :+1:

PR #1686, opened 2016-08-08
Load tether

andrewklau:


chrisatomix:
Why do we need Tether in Sage?

allanchau:
It's used for the Bootstrap tooltips in v4 and already included in the sage 9 branch.

andrewklau:
Yes, required for tooltips in bootstrap v4. This line is needed or it will complain "Tether" is undefined.

PR #1696, opened 2016-08-26
Set dynamically aboslute public path on 'npm run watch'; fix http/htt…

ptrckvzn:
A new proposition to set public-path dynamically when runnning `npm run watch`. Simpler config and watch.js that fixes http/https.

This addresses the issues in https://github.com/roots/sage/pull/1661.

### DIFF
--- a/assets/config.json
+++ b/assets/config.json
@@ -2,6 +2,7 @@
   "context": "assets",
   "entry": {
     "main": [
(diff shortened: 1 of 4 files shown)


retlehs:
thanks! if you rebase we can merge

ptrckvzn:
Ok, done.

PR #1703, opened 2016-09-01
Refactor build routine (WIP)

QWp6t:
**Update:** Apparently this is not compatible with anything lower than Node 6.x. I'll have to take care of that.

Right now this should work. You should be able to run the following commands out of the box.

```
npm start
npm run build
npm run build:production
npm run clean
npm test
```
`npm start` and `npm test` are aliases for `npm run watch` and `npm run lint` (although `test` could be modified to run actual tests as well, obviously).

I haven't done much testing, and the commits are a little messy. I'll likely rebase it and add some better commit messages before merging.


ptrckvzn:
That's cool ! I'm looking forward testing and using this !

Two things (I tested no. 1, but not no. 2):

1. 'public-path.js' should be added to every [entry](https://github.com/roots/sage/pull/1703/files#diff-1811fd0f74b81c5fa5a7a330d7d767cbR15) in assets/config.js, otherwise css url() won't load properly when running the watcher.

2. shouldn't the loader for [node_modules|bower_components](https://github.com/roots/sage/pull/1703/files#diff-fa60332fc10c5c5204e38b00eaf0bf6cR103) images and fonts add a [hash] to the filenames at all time? Since it doesn't replicate the folder tree and put everything in dist/vendor/ there is a slim chance of collision for two assets with the same name.

PR #1706, opened 2016-09-02
Update _grid.scss to use @include make-col-ready() mixin

darrenjacoby:
Rather @include make-col-ready() as opposed to using @extend %grid-column for classes main and sidebar.
http://v4-alpha.getbootstrap.com/layout/grid/


QWp6t:
Thanks!

PR #1712, opened 2016-09-05
Namespaces on own line

kalenjohnson:
namespace declaration on the same line as <?php isn't being used much anymore. Both PHP FIG and Laravel show the namespace on it's own line.

Reference:
http://www.php-fig.org/psr/psr-2/
https://github.com/laravel/laravel/blob/master/app/User.php#L1-L3

### DIFF
--- a/src/admin.php
+++ b/src/admin.php
@@ -1,4 +1,6 @@
-<?php namespace App;
+<?php
+
+namespace App;
 
 /**
  * Theme customizer
--- a/src/filters.php
+++ b/src/filters.php
@@ -1,4 +1,6 @@
-<?php namespace App;
+<?php
+
+namespace App;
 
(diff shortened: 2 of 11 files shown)


QWp6t:
It's still PSR-2 (and PSR-1) compliant, though, to be clear. PSR-2 doesn't require or even suggest that namespaces be on their own line or that `<?php` be on its own line.

I'm cool either way as far as this goes. I personally like the look of `<?php namespace ...` more but it's not a big deal.

kalenjohnson:
You're right, I thought PSR-2 did say that namespace needs to be on it's own line, but it doesn't, only that it requires a line afterwords.

I personally just haven't seen the namespace on the first line in a long while, threw me off :joy: 

I'm open to either as well

PR #1733, opened 2016-10-05
Fix duplicate value in Browsersync files Array

ptrckvzn:
The two same values get concatenated. A console.log output of `watcherConfig` in `webpack.plugin.browsersync.js` : 

```
{ 
  host: 'localhost',
  port: '3000',
  proxy: 
   { target: 'https://example.dev',
     middleware: [ [Object], [Object] ] },
  files: 
   [ 'templates/**/*.php',
     'src/**/*.php',
     'templates/**/*.php',
     'src/**/*.php' ]
}
```

### DIFF
--- a/assets/build/config.js
+++ b/assets/build/config.js
@@ -27,10 +27,7 @@ const config = mergeWithConcat({
     watcher: !!argv.watch,
     uglifyJs: !(argv.p || argv.optimizeMinimize),
   },
-  watch: [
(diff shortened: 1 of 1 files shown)


QWp6t:
Yeah, I fixed this last week locally, but I haven't pushed it yet since there are still a few other things I'm working on. Thanks for the PR.

PR #1737, opened 2016-10-14
Use Bedrock-style folder structure on publicPath by default

villeristi:
Use Bedrock-style folder structure on webpack `publicPath` by default, yet still overridable via `config.json`.


QWp6t:
The last time this was brought up, we discussed it and decided that that users have to edit the `devUrl` anyway, so it's not an issue for them to just edit the `publicPath` while they're at it.

PR #1741, opened 2016-10-24
Ensuring the execution of 'Router.loadEvents'

kaisermann:
Sometimes the `DOMContentLoaded` event has already been fired. To make sure the `loadEvents()` method is always executed, we should check the `document.readyState` and, if it’s already defined as "complete", execute the callback immediately.

[Reference](https://github.com/jquery/jquery/blob/ad6a94c3f1747829082b85fd53ee2efbae879707/src/core/ready.js#L80-L93)


retlehs:
closed by #1751

PR #1744, opened 2016-10-31
Adds Yarn package and lock file.

raminv80:
Adds Yarn to speed up installation and increase stability and security.


chrisatomix:
Why would you want Yarn installed locally? It should be installed globally, ie:
`npm install -g gulp bower yarn`

QWp6t:
Just add the yarn lock file. There's no reason to have yarn itself as a dependency. And if it was going to be a dependency, it should be a devDependency. But again, there's no need to list it at all. Just having a yarn.lock file is sufficient.

raminv80:
Yarn was removed from package dependency list in favor of its global installation. The PR now only includes the `yarn.lock` file.

PR #1748, opened 2016-11-02
Remove gulp raster

jdonisvitch:


jdonisvitch:
At this point I'm so dumb I've had to rename the fork i've made, thanks again to all the maintainers!

Issue #1715, opened 2016-09-11
[bug] $(document).ready() on development env


jmarceli:
Hi.
When hot reloading is enabled `ready` event is triggered before CSS styles are applied. I run into this issue when I've tried to determine active media query based on current CSS style. It works in production/release but doesn't work with webpack.

The only workaround which I manage to came up with is defining `'process.env'` variable in the `webpack.config.js` file and based on its value switching `$(document).ready()` to `$(window).load()`.

Here is the code:
**assets/build/webpack.config.js**

```
    new webpack.DefinePlugin({
      WEBPACK_PUBLIC_PATH: (argv.watch === true) ? JSON.stringify(path.join(config.publicPath, dist)) : false,
      'process.env': argv.release ? "'production'" : "'development'"
    })
```

**assets/scripts/main.js**

```
// Load Events
if (process.env === 'production') {
  $(document).ready(() => new Router(routes).loadEvents());
} else {
  $(window).load(() => new Router(routes).loadEvents());
}
```

Please tell me if there is any cleaner solution.

**As a side note**: I think that some kind of `process.env` variable might be useful in other cases too. Maybe it is a good idea to provide it by default?

QWp6t:
Looks like you're using alpha 1. This might have been resolved in alpha 2 or in the master branch. Feel free to post here if it's still a problem in the newer releases, and we can revisit this issue.

jmarceli:
Hi, I've adapted code from alpha 3 which is:
```
```

QWp6t:
Okay, I'll look into it.

In the meantime, you can use `module.hot` instead of `process.env`.

https://webpack.github.io/docs/hot-module-replacement.html

jmarceli:
One more detail regarding this issue. The function which has inconsistent behaviour looks as follow:
```
  getBreakpointName() {
    const breakpoint = window.getComputedStyle(
      document.getElementById('test'), ':before'
      ).getPropertyValue('content').replace(/\"/g, '');
    return breakpoint;
  }
```
So I basically returns the `content` of selected DOM element `:before` pseudo selector.

retlehs:
closed by #1751

Issue #1743, opened 2016-10-28
ACF Local JSON feature broken in Sage 9


samrap:
## Submit a feature request or bug report
- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

Replace any `X` with your information.

---

**What is the current behavior?**

When attempting to use the [ACF](https://www.advancedcustomfields.com) plugin's [Local JSON](https://www.advancedcustomfields.com/resources/local-json/) feature, ACF JSON data is never saved into the `acf-json` directory.

**What is the expected or desired behavior?**

Expected behavior is what is outlined in the ACF docs for [Local JSON](https://www.advancedcustomfields.com/resources/local-json/).

---
## Bug report

(delete this section if not applicable)

**Please provide steps to reproduce, including full log output:**
- Install ACF plugin
- Create `acf-json` directory in the theme folder.
- Attempt to create or modify a custom field

At this point, a JSON file should be saved into the `acf-json` directory. Due to how Sage 9 rewrites the theme location, this does not happen.

**Please describe your local environment:**

WordPress version: 4.6

OS: OSX & Linux

NPM/Node version: N/A

**Where did the bug happen? Development or remote servers?**

Local, Virtual, and Production servers.

---

**Other relevant information:**

I published a Gist that fixes the issue. I was hesitant to mark it as a bug, since not everyone uses ACF. However, it is recommended by Sage and thus it may be beneficial to add this fix into the Sage `setup.php` before the official release, or at least documenting the issue. If this is something the Sage developers would like to fix, I would be more than happy to submit a PR!

In the meantime, here is the Gist for any puzzled Sage developers out there: https://gist.github.com/samrap/d4a19858c2ed1d238c27df91580350e0

AdamWills:
Definitely good to know, as we use this feature on almost every site we build.

retlehs:
[...]

Issue #1742, opened 2016-10-27
Autoprefixer & postcss plugins on postcss loader


wab:
## Submit a feature request or bug report
- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Autoprefixer (or any postcss plugin like postcssFlexbugsFixes) seems not to proceed on scss files.

**What is the expected or desired behavior?**

Prefix or plugin result on output css file

---
## Bug report

**Please provide steps to reproduce, including full log output:**

with loader 

``` javascript
{
        test: /\.scss$/,
        include: config.paths.assets,
        loader: ExtractTextPlugin.extract({
          fallbackLoader: 'style',
          loader: [
            `css?${sourceMapQueryStr}`,
            'postcss-loader',
            `resolve-url?${sourceMapQueryStr}`,
            `sass?${sourceMapQueryStr}`,
          ],
        }),
      },
```

and this config : 

``` javascript
new webpack.LoaderOptionsPlugin({
      minimize: config.enabled.minify,
      debug: config.enabled.watcher,
      stats: {
        colors: true,
      },
      postcss: [
        autoprefixer({
          add: true,
          browsers: [
            'last 2 versions',
            'android 4',
            'opera 12',
          ],
        }),
      ],
      eslint: {
        failOnWarning: false,
        failOnError: true,
      },
 }),
```

for example a simple main.scss entry like

``` css
@keyframes move {
  from { top: 0; left: 0; }
  to   { top: 100px; left: 100px; }
}

.move {
  animation: move 10s steps(10) infinite alternate;
}

.flex {
  display: flex;
  > * {
    flex: 1;
  }
}

.fs {
  &:fullscreen {
    width: 100vw;
    height: 100vh;
  }
}
```

 output to main.css

``` css
@keyframes move {
  from {
    top: 0;
    left: 0;
  }

  to {
[...]

PR #1739, opened 2016-10-19
Suppress chunk messages

drawcard:
Addresses the discussion in https://github.com/roots/sage/issues/1719

PR #1751, opened 2016-11-07
Sage 9-alpha.4

QWp6t:
I haven't done a whole lot of testing with this so I don't wanna push it as an update just yet. 

Will do some more tests and tag a new release probably sometime this week. If all goes well with this release, then the next step is adding Blade support.

## If you decide to test this, be sure to re-activate your theme.


QWp6t:
This PR closes: #1715, #1719, #1739, #1741, #1742, #1743

nathanielks:
Just tested this on a theme I'm working on. Javascript is loading, but for some reason styles aren't being compiled?

JonMasterson:
I encountered an issue related to the 'stylesheet' > 'template' revision in functions.php (not using Bedrock). With the 'template' change, theme name was missing from my css/js paths. Also, image assets went missing — image file paths (which should point to '/dist/images') contained '/templates/' in the file path. When I reverted functions.php, all was well.

QWp6t:
@nathanielks 
If you mean they're not written to your filesystem when you do `npm start`, that's because they're in webpack's internal filesystem, which sits in memory. This is faster. When you visit the proxy server, they should still be served.

@JonMasterson 
You just have to reactivate your theme. That will solve your issue.

nathanielks:
@qwp6t yup, I'm aware! The stylesheets weren't loaded even when visiting the proxy. I'll test a vanilla clone and see what that gives me

JonMasterson:
Doh! Didn't notice the hook — Works great. Thanks!

giacomoalonzi:
@nathanielks same problem here, let me know if you solve it! thx

retlehs:
@giacomoalonzi try removing your `node_modules` directory and switching to yarn

nathanielks:
@retlehs @giacomoalonzi I was already using `yarn` and nuked `node_modules` to no avail... an interesting clue is the output for `yarn start` in alpha4 for me does not include any of the standard Webpack jargon and doesn't appear to run anything when I save a file that _would_ trigger an update in alpha3.

nathanielks:
Mmmm'k. Vanilla alpha4 _does_ have the webpack jargon, so I'm believing I didn't merge everything correctly. I had a bunch of merge conflicts trying to merge 3 and 4 in an existing theme, so it was a bunch of manual merge resolution.

nathanielks:
Welp, insert foot into mouth. It works fine! The only difference for me is the Webpack hullabuhloo still isn't being output in `yarn start`, but it does in fact perform updates in the browser.

nathanielks:
Now, one thing that I have noticed (that's semi annoying) is that on subsequent rebuilds without refreshing the page, updated JS isn't really because the router is only loaded on document.ready() (this also existed in alpha3). @QWp6t is the solution here to have it trigger on document.ready and on rebuilds? Does HMR broadcast any events we can hook into? I'm just not sure where to start. I imagine we'll have to have some method of removing event listeners as well before the code is evaluated...

nathanielks:
Oh, and for whatever reason even reloading the page doesn't update the js. I have to `ctrl-c` `yarn start` and restart it to get the updates in the browser.

QWp6t:
@nathanielks 

I'm not sure if there's an event, but this should work:
```
if (`module.hot`) {
  new Router(routes).loadEvents();
}
```

Obviously an event would be better, though.

derkjn:
Side note, this is already not compatible with `webpack@2.1.0-beta.26`.
The `package.json` should be updated accordingly.

wassimdotis:
I can't get isotope and a couple of other scripts to work with DOM ready. What was the reason for 724d550739c73a561ab80f9d12add89fb78a7c8c ?

derkjn:
Tested this today on a fairly big project with lots of other deps and stuff. All working good on my end! Cheers @QWp6t!

Issue #1699, opened 2016-08-30
Add support for Laravel Blade


QWp6t:
I've come up with a couple ways to implement this, but I'm still undecided on the approach I'm going to take and I haven't fully investigated whether there are any pitfalls. Definitely need to do more testing and research, but it only took me about 10 minutes to get it working so I anticipate it being fairly straightforward.

nathanielks:
```
@section('comment')
    @each('lists.awesome-item', $reasons_quap_is_awesome, 'reason')
@endsection
```
_Renders:_
1. Quap is awesome because: he thinks outside of the box
1. Quap is awesome because: he's kindler and gentler
1. Quap is awesome because: he contributes to FOSS

PR #1765, opened 2016-12-03
Blade

kalenjohnson:
Blade templating. This PR is going to be dependent on WP 4.7, specifically [this](https://github.com/WordPress/WordPress/blob/4.7-branch/wp-includes/template.php#L29-L41) section of code. If you'd like to test, either use WP 4.7 or grab this code and add it to your installation. If you're using Bedrock, you can set the version requirement of `johnpbloch/wordpress` to `4.7.x-dev`.

Currently, `.blade.php` files are used, and they are being added to the list of files to load via the previously mentioned new hook.

For plugins, the same issues I believe will apply that have been in Sage for the last few versions - plugins are able to load their own templates, but the same issues with calls to `get_header()` and `get_footer()` will still show deprecation warnings. You'll want to add a Blade template according to the WP template you need to override.

Looking for feedback and opinions on the current setup.

Ref #1699


oxyc:
Wouldn't this fit well as a soil module? And then sage or other themes can opt in to use blade templating (or timber which I'm using).

corradomatt:
I also like the idea of it being a soil module which allows for a user to choose between Timber or Blade.

christianmagill:
Since you are reconstructing the templates array, could it be made available elsewhere so for instance a header or sidebar could be retrieved based on the first match on the template hierarchy?

retlehs:
this won't be a soil module, partly because soil is a paid plugin and partly because we don't think it fits in there

@kalenjohnson has suggested we create a composer package for it within this repo, similar to how laravel does in [all of these subfolders](https://github.com/laravel/framework/tree/5.3/src/Illuminate)

src/filters.php
```diff
@@ -32,11 +33,62 @@
...
+    return false;
+}, 1000);
+
+add_filter('comments_template', function ($theme_template) {
+    $container = Container::getInstance();
+    $template_engine = $container->make('view');
+
+    $template_name = basename(str_replace('.blade.php', '', $theme_template));
+    $html = $template_engine->make('partials/'.$template_name, []);
+    var_dump($template_engine->exists('partials/'.$template_name));
```
aaemnnosttv: Oops. ;)

src/filters.php
```diff
@@ -32,11 +33,62 @@
...
+]);
+add_filter('template_include', function ($template) {
+    $blade_template = (!strpos($template, '.blade.php')) ? str_replace('.php', '.blade.php', $template) : $template;
+    $blade_template = locate_template(basename($blade_template));
+
+    if (!file_exists($blade_template)) {
+        return $template;
+    }
+
+    $container = Container::getInstance();
```
kalenjohnson: I can understand the reasoning on this, making sure everything that's used is in the same namespace. However, I was also considering that really, the Container class should be global and only one instance of, just like in a Laravel application. The IoC container is something that can be used anywhere and give you instances of anything you need.

Basically, I don't know if we necessarily want to make that Roots specific. I don't think it would necessarily be bad if other plugins called `Illuminate\Container\Container` directly. On the other hand, perhaps we do move it out to it's own project along with the Blade service provider, and also add some things that are used in the global scope of WP, like `$wpdb`, and add that to the container, it's a container that could be used across plugins and themes in general.

QWp6t:
Also added `@asset()` Blade directive

It uses `App\asset_path()` to grab cache-busted URL

schrapel:
Thanks so much for your work on this! What do you think about creating some custom directives for the typical use cases such as `have_posts()`. There is quite a lot of php just hidden in blade syntax.  

https://laravel.com/docs/5.3/blade#extending-blade

christianmagill:
Can't wait to try this out! Would it be possible to allow injecting data into the views? I see that the template method accepts data but it doesn't appear to be used in the render. Perhaps a filter could be used.

robin-dwi:
Awesome work! I just love the blade engine and can't wait to start building stuff :)

Did a quick test run and it seems like the `page.blade.php` loads even though I select `template-custom.blade.php`.

Yes, I did make changes to the custom template.

src/filters.php
```diff
@@ -32,11 +29,28 @@
...
-    }
-    return ((new Template(new Wrapper($main)))->layout());
-}, 109);
+array_map(function ($tag) {
+    add_filter("{$tag}_template_hierarchy", function ($templates) {
+        return array_merge(str_replace('.php', '.blade.php', $templates), $templates);
+    });
+}, [
+    'index', '404', 'archive', 'author', 'category', 'tag', 'taxonomy', 'date', 'home',
+    'front_page', 'page', 'paged', 'search', 'single', 'singular', 'attachment'
```
davekiss: `front_page` should be corrected to `frontpage` https://github.com/WordPress/WordPress/blob/4.7-branch/wp-includes/template.php#L29-L41

QWp6t: You can submit PR if you want credit for it. Otherwise I'll handle it later.

PR #1767, opened 2016-12-07
Tighten version constraint for illuminate/view

aaemnnosttv:
The version constraint for `illuminate/view` should probably be tighter, as Laravel follows their own flavor of Semver in which breaking changes are allowed between minor releases.

This could potentially create a situation where simply running `composer update` could break the site (eg. after 5.4 is released).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
   "require": {
     "php": ">=5.6.4",
     "composer/installers": "~1.0",
-    "illuminate/view": "^5.3",
+    "illuminate/view": "~5.3.0",
     "jenssegers/blade": "dev-master#59ba2cc"
   },
   "require-dev": {


QWp6t:
Out of curiosity, do you know what breaking changes (or how many) Laravel has introduced into `illuminate/view` since 5.0?

aaemnnosttv:
I just took a quick look through the [upgrade guide](https://laravel.com/docs/5.3/upgrade), and I didn't see any breaking changes for the view package specifically, but keep in mind that [`illuminate/view`](https://packagist.org/packages/illuminate/view) requires 5 other illuminate packages, including `illuminate/support`.

I bring this up because I was bit by this recently when 5.3 was released from breaking changes within the `Illuminate\Support\Collection` class.

[Laravel](https://packagist.org/packages/laravel/laravel) itself currently requires the framework using `5.3.*`.

I just think it's better to follow "the Laravel way" in this case and be slightly more conservative with the version constraint. It can always be bumped later.

QWp6t:
Sounds good. Thanks!

PR #1772, opened 2016-12-12
front_page => frontpage

davekiss:
Correct the value based on what WordPress passes as filter values

https://github.com/WordPress/WordPress/blob/4.7-branch/wp-includes/template.php#L29-L41


retlehs:
thanks!

QWp6t:
wo that was fast

PR #1782, opened 2016-12-28
Throw errors if minimum requirements aren't met

QWp6t:
Checks for all requirements and throws an error if one isn't met.
* Minimum PHP version (`>= 5.6.4`)
* Autoloader (from Sage directory, Bedrock, or whatever; just has to be able to access Container class)
* Required files
  * helpers.php
  * setup.php
  * filters.php
  * admin.php

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -5,23 +5,46 @@
  */
 
 /**
- * Require Composer autoloader if installed on it's own
+ * Helper function for prettying up errors
+ * @param string $message
+ * @param string $subtitle
+ * @param string $title
  */
-if (file_exists($composer = __DIR__ . '/vendor/autoload.php')) {
-    require_once $composer;
+$error = function ($message, $subtitle = '', $title = '') {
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -5,23 +5,46 @@
...
+    $footer = '<a href="https://roots.io/sage/docs">roots.io/sage/docs</a>';
+    $message = "<h1>{$title}<br><small>{$subtitle}</small></h1><p>{$message}</p><p>{$footer}</p>";
+    wp_die($message, $title);
+};
+
+/**
+ * Ensure compatible version of PHP is used
+ */
+if (version_compare('5.6.4', phpversion(), '>=')) {
+    $error(__('You must be using PHP 5.6.4 or greater.', 'sage'), __('Invalid PHP version', 'sage'));
```
swalkinshaw: "PHP 5.6.4 or greater is required to use Sage." maybe?

PR #1792, opened 2017-01-08
Update to Bootstrap 4 Alpha 6

retlehs:
this also drops the flexbox option we previously enabled, since bs4 is now flexbox by default (and without the option to disable)

http://blog.getbootstrap.com/2017/01/06/bootstrap-4-alpha-6/

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -422,12 +422,12 @@ boom@2.x.x:
   dependencies:
     hoek "2.x.x"
 
-bootstrap@^4.0.0-alpha.5:
-  version "4.0.0-alpha.5"
-  resolved "https://registry.yarnpkg.com/bootstrap/-/bootstrap-4.0.0-alpha.5.tgz#a126b648c3bd2f52b8fad4bbc5e2d0ad2abf7064"
+bootstrap@^4.0.0-alpha.6:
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -422,12 +422,12 @@ boom@2.x.x:
...
-bootstrap@^4.0.0-alpha.5:
-  version "4.0.0-alpha.5"
-  resolved "https://registry.yarnpkg.com/bootstrap/-/bootstrap-4.0.0-alpha.5.tgz#a126b648c3bd2f52b8fad4bbc5e2d0ad2abf7064"
+bootstrap@^4.0.0-alpha.6:
+  version "4.0.0-alpha.6"
+  resolved "https://registry.yarnpkg.com/bootstrap/-/bootstrap-4.0.0-alpha.6.tgz#4f54dd33ac0deac3b28407bc2df7ec608869c9c8"
   dependencies:
-    jquery "1.9.1 - 3"
-    tether "^1.3.7"
+    jquery ">=1.9.1"
```
QWp6t: We could possibly use `>=` in Sage's package.json too, but afaik jQuery has only guaranteed API compat up to v3, which is the reason for the 1-3 range.

Issue #1794, opened 2017-01-08
Images added to template during Browsersync session not accessible


retlehs:
if you add an image to `assets/images/` during a browsersync session it isn't yet accessible from the templates. works fine within styles.

related topic:

https://discourse.roots.io/t/webpack-watch-images-in-assets-images-not-copied-to-dist/8487

Issue #1790, opened 2017-01-08
JavaScript changes require a page refresh within Browsersync session


retlehs:
we're aware of this issue, wanted to raise it on here so that y'all know we are working on a solution.

related topic:

https://discourse.roots.io/t/sage-9-js-doesnt-update-on-reload/8474

andrewmartin:
Hi there, I'm hitting this issue still. I've tried following what was going on in that post, but I'm hitting a really annoying refresh loop if I make any changes to JS. Any suggestions on how to fix?

andrewmartin:
Strangely, I changed

`"publicPath": "/app/themes/mysite",`

to

`"publicPath": "/wp-content/themes/mysite",`

And it seems to have fixed the problem. Thought I'd bring this up because I just used the default generator and didn't touch `config.json`, did I do some step wrong when I went through the CLI?

retlehs:
the cli asked:

> Path to theme directory (e.g., /wp-content/themes/themename)

and instead of entering in the correct path it sounds like you accepted the default

by default sage uses `/app/themes/sage` as the `publicPath` as we're assuming you're using a bedrock install

please use https://discourse.roots.io/ for support

pmascis:
I too was receiving the refresh loop @andrewmartin mentioned when changing any JS.
I accidentally had a trailing slash at the end of my theme name, which was causing the same thing.
i.e.
"publicPath": "/wp-content/themes/mysite/"
Instead of
"publicPath": "/wp-content/themes/mysite"

Sorry to revive this, just thought I'd mention in case anyone else makes the same silly mistake.

builtbyvern:
I don't have a trailing slash and deal with this issue. It works 90% of the time but about twice an hour I get stuck in a reload loop where I have to literally empty out every file I have, hard refresh like 10x and then resave every file til it works again.

knowler:
@builtbyvern 

I’ve collected a bunch of checks/fixes/solutions to Browsersync related issues on Discourse:
https://discourse.roots.io/t/sage-9-browsersync-not-updating-right/10648/11?u=knowler

The [Roots Discourse](https://discourse.roots.io) is the best place to find support.

Issue #1789, opened 2017-01-08
publicPath should support multiple environments


retlehs:
https://github.com/roots/sage/blob/e6d187cb6a1778de907b38a7ee8708745d78714d/assets/config.json#L15

sometimes this path on dev will end up being different on production. (eg. you're using bedrock on local dev, but unfortunately have to deploy to a vanilla wordpress installation :cry:)

@QWp6t also mentioned possibly using `../` as a solution

related topics:

https://discourse.roots.io/t/publicpath-dev-vs-prod/7623
https://discourse.roots.io/t/deploy-with-capistrano-wrong-assets-path-in-webpack/8458

PR #1795, opened 2017-01-10
Beta1

QWp6t:
Updated dependencies and resolved some issues with file watching, browsersync, and publicPath.


QWp6t:
#1794, #1790, #1789 should be good now but require documentation

PR #1796, opened 2017-01-10
Use stock eslint

swalkinshaw:


swalkinshaw:
We can iterate on this and define a more styling rules if we want.

PR #1798, opened 2017-01-10
Allow browser versions to be configured in config.json

oxyc:
As sage is a starter theme it shouldn't make hardcoded assumptions about which browsers are supported. This PR allows users to override the default browser array passed to autoprefixer. It also passes this array on to cssnano to avoid stripping intentional vendor prefixes.

### DIFF
--- a/assets/build/config.js
+++ b/assets/build/config.js
@@ -26,6 +26,7 @@ const config = mergeWithConcat({
     watcher: !!argv.watch,
   },
   watch: [],
+  browsers: [],
 }, userConfig);
 
 config.watch.push(`${path.basename(config.paths.assets)}/${config.copy}`);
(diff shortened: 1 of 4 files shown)


assets/build/config.js
```diff
@@ -26,6 +26,7 @@ const config = mergeWithConcat({
     watcher: !!argv.watch,
   },
   watch: [],
+  browsers: ['last 2 versions', 'android 4', 'opera 12'],
```
QWp6t: Can you make this line:
```json
  browsers: []
```

And add those browser versions to `config.json` instead?

The reason is because arrays are concatenated by `mergeWithConcat()`.

retlehs:
thank you!

PR #1809, opened 2017-01-13
Remove Font Awesome

retlehs:
font awesome was added in sage 9 mainly to provide a good example out of the box on how to add a 3rd party package

there's a few issues:

* not everyone wants it
* font awesome includes formats that aren't necessary right now (https://github.com/FortAwesome/Font-Awesome/issues/10470)
* produces webpack warnings about the entrypoint size limit out of the box

this will instead be moved to the docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### 9.0.0-beta.2: January 11th, 2016
+* Remove Font Awesome ([#1809](https://github.com/roots/sage/pull/1809))
 * Remove grid defaults ([#1808](https://github.com/roots/sage/pull/1808))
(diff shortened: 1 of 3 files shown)


MWDelaney:
What would be nice is good documentation on how to add it again. This and the default grid (removed in Beta 2) provided really good examples to work from especially considering that 1) "how do I add FontAwesome" is a REALLY frequent question on Discourse, and 2) Bootstrap 4 doesn't (yet?) effectively document how to use 
```
  @include make-col-ready();
  @include media-breakpoint-up(sm) {
        ...
   }
```

So I frequently reference `_grid.scss`.

These examples could be easily maintained on a page on the website, though.

PR #1815, opened 2017-01-19
Fix browsersync

QWp6t:
I believe the main issue was `monkey-hot-loader`, which I just removed since the original repo has been abandoned and the fork we were using didn't seem like it was likely to be maintained either.

Just read commit messages for other changes that were made.


kalenjohnson:
We won't read commit messages, we demand you explain yourself.

We the people

PR #1817, opened 2017-01-20
Add support for HTML injection

### LINKED ISSUE #1816: Browsersync doesn't detect any change to any php file

- [ ] This is a feature request
- [x] This is a bug report

---

**What is the current behavior?**

After the merge with `beta.2`, `Browsersync` doesn't detect any change to the php files either in `src` nor in `templates` folders.

**What is the expected or desired behavior?**

As before the latest update, `Browsersync` should automatically refresh the window when the templates or classes in the project are updated

---

## Bug report

**Please provide steps to reproduce, including full log output:**

Install a fresh copy of the theme.
Run `yarn run start`
Start changing your templates, you'll notice nothing happens and you have to manually refresh your browser.

**Please describe your local environment:**

WordPress version: 4.7.1

OS: 10.12.2

NPM/Node version: 3.10.8/6.9.1

**Where did the bug happen? Development or remote servers?**

Development

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

No

QWp6t:
Resolves #1816


assets/build/webpack.config.watch.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
       target: config.devUrl,
       proxyUrl: config.proxyUrl,
       watch: config.watch,
+      delay: 500,
```
oxyc: In the case you're editing a file that causes a refresh and you save the file multiple times, there's the possibility that you get an outdated version if the browser is in the middle of a refresh when the save is detected.

I ran into this and used [`reloadDebounce`](https://browsersync.io/docs/options/#option-reloadDebounce) instead. Seems pointless to trigger multiple refreshes in a row anyways.

QWp6t: The delay is an arbitrary one set by my plugin to invoke a delay between local file system updates and injection/reload by BrowserSync. This is to resolve an issue with BrowserSync reloading before the server (i.e., VM) has synchronized with local file changes. My plugin by default has this value set to 50ms, but it's 500ms in Sage because I'm assuming that most developers are using a VM.

Anyway, I could add a separate option for debounce as well, or possibly refactor my current implementation to use debounce instead. I'll look into it and do some tests.

QWp6t:
@oxyc I added a debounce option to the plugin so you can set `debounce` in webpack.config.watch.js if you want, and it will get passed as the value to `reloadDebounce`. I might wind up implementing my own debounce as well to handle the different ways of resolving file update events.

PR #1821, opened 2017-01-26
Update functions.php

schrapel:
I don't propose this as the solution, it's kind of ugly but...

If you load the theme from your main composer file e.g. the Bedrock one. The class will exist so the if statement above fails and it'll try to require a file that doesn't exist

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -27,13 +27,15 @@ if (version_compare('5.6.4', phpversion(), '>=')) {
 /**
  * Ensure dependencies are loaded
  */
-if (!file_exists($composer = __DIR__.'/vendor/autoload.php') && !class_exists('Roots\\Sage\\Container')) {
-    $sage_error(
-        __('You must run <code>composer install</code> from the Sage directory.', 'sage'),
-        __('Autoloader not found.', 'sage')
-    );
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -34,7 +34,7 @@
     );
 }
 
-if ($composer) {
+if (file_exists($composer)) {
```
mAAdhaTTah: Instead of doing this again, you can save the result of the `file_exists` check above and reuse that value here. Some extra parens wrapping may be required, but that's how I resolved this problem in a theme I'm working on that uses a couple of Sage concepts.

schrapel: Yeah, that's why I said the solution was ugly. Just wanted to raise it as an issue

QWp6t:
```php
/**
 * Ensure dependencies are loaded
 */
if (!class_exists('Roots\\Sage\\Container')) {
    if (!file_exists($composer = __DIR__.'/vendor/autoload.php')) {
        $sage_error(
            __('You must run <code>composer install</code> from the Sage directory.', 'sage'),
            __('Autoloader not found.', 'sage')
        );
    }
    require_once $composer;
}
```

PR #1828, opened 2017-02-02
Add babel-eslint as default parser

gretzky:
Taken from [here](https://github.com/yannickcr/eslint-plugin-react/issues/447#issuecomment-184617282)

Was playing around with Sage and ran into an eslint error-
`Parsing error: The keyword 'import' is reserved`

Discovered that my eslint was out of date on my local machine, did some digging and came across the thread above. Adding babel-eslint as the parser fixed it. Might save some people headaches and keep consistency.

### DIFF
--- a/assets/build/.eslintrc
+++ b/assets/build/.eslintrc
@@ -1,5 +1,6 @@
 {
   "root": false,
+  "parser": "babel-eslint",
   "rules": {
     "import/no-extraneous-dependencies": 0,
     "prefer-rest-params": 0,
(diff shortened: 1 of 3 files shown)


QWp6t:
I don't think this is necessary.

What version of node are you using? What version of npm/yarn? Your globally installed version of eslint shouldn't matter since it's installed locally via Sage's package.json.

eslint supports ECMAScript 3. 4, 5, 6, 7, and 8 out of the box. By default it's set to ECMAScript 5, but [we set it to 8](https://github.com/roots/sage/blob/master/.eslintrc#L21).

It sounds like this was just configuration issue on your end; or is there something else I'm missing here? Feel free to provide me with a reduced test case that reproduces the issue you're having, and I'll reopen this PR.

PR #1814, opened 2017-01-18
Bootstrap 4 alpha 6 dependency update, error recovery for gulp.

MWDelaney:
Second attempt at this PR after a bad push ate my 8.5.0 branch. See [here](https://discourse.roots.io/t/jshint-crashing-gulp-when-a-javascript-error-occurs/6783/8) for the error recovery code (thanks to Nicolo_Sacchi and epiphron!). I didn't write it, I just turned it into a PR.

PR #1832, opened 2017-02-07
Fix gulp error handling (Sage 8)

TobiasKjrsgaard:
This PR is basically just a rebased version of #1814 with a few whitespace and indentation fixes, sqashed down to two commits. It addresses the `gulp watch` issue mentioned on [Discourse](https://discourse.roots.io/t/jshint-crashing-gulp-when-a-javascript-error-occurs/6783).

It also updates Bootstrap to version 4.0.0-alpha.6 and removes the `$enable-flex` var.

### DIFF
--- a/assets/styles/common/_variables.scss
+++ b/assets/styles/common/_variables.scss
@@ -1,5 +1,4 @@
 // Grid settings
-$enable-flex:           true;
 $main-sm-columns:       12;
 $sidebar-sm-columns:    4;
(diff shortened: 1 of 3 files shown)


retlehs:
thanks! i'll tag 8.5.1 later

theatrain:
Any reason why you would use BS4 which is in alpha as part of a production release?

PR #1834, opened 2017-02-09
Add Github URI Theme Location

drawcard:
This proposed change adds information to allow simple updating the of the Sage theme, using the Github Updater plugin for WP (https://github.com/afragen/github-updater). It allows the theme to be managed and updated natively in the Wordpress backend as changes are made. Could be very useful for parent themes to stay current / up to date.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,6 +1,8 @@
 /*
 Theme Name:         Sage Starter Theme
 Theme URI:          https://roots.io/sage/
+GitHub Theme URI:   roots/sage
+
 Description:        Sage is a WordPress starter theme.
 Version:            9.0.0-beta.2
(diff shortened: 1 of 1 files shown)


aaemnnosttv:
I feel like this would cause more problems than the benefit it aims to provide.  Sage is a starter theme, designed to be changed rather than extended as a parent theme.

Adding this makes such updates opt-out rather than opt-in, and could could easily create the possibility of someone's custom Sage being completely wiped out unintentionally by a new release.

Also, Sage has major backwards-incompatible changes between major versions, something which automatic updates via WP or GitHub Updater do not protect against.

drawcard:
Fair enough, I see your point :)

PR #1835, opened 2017-02-14
Change path in asset_path() function

darkwebside:
Instead using get_template_directory() use get_stylesheet_directory_uri(). because it doesn't find main.css and main.js


retlehs:
sage isn't built with child themes in mind out of the box & we're no longer really making changes to the 8 branch anymore

PR #1839, opened 2017-02-21
Fix gulp imagemin options

gnowland:
Image minification options (specifically SVG plugin options) were not
working when using the latest version of gulp-imagemin due to a new
options passing structure, see:
https://github.com/sindresorhus/gulp-imagemin/releases/tag/v3.0.0


retlehs:
thanks!

PR #1849, opened 2017-02-27
Check for dynamic env DEVURL before config

slackday:
This makes it possible to watch files on a different URL than what is configured in `assets/config.json`.

I find this useful when working with a Multisite install on sub-domains and want to watch while working on a sub-site.

Or when working together with developers that have a different LAMP stack than Trellis. This doesn't require changing and commiting changes to `assets/config.json` simply start the process with:

`DEVURL=https://subsite.example.dev npm start`


QWp6t:
Dig the idea. :+1:

@retlehs Thoughts?

retlehs:
thanks!

PR #1852, opened 2017-03-01
Add version check for WordPress version less than 4.7.0

rfair404:
Sage checks if the version of PHP installed is less than 5.6. However, if the installed version of WordPress is less than 4.7, a fatal error is thrown. This check produces a more friendly "Sage Style" error instead of a boring default php fatal error. 

See: https://discourse.roots.io/t/undefined-function-app-get-theme-file-path/8455

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -25,6 +25,13 @@ if (version_compare('5.6.4', phpversion(), '>=')) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


rfair404:
@JulienMelissas

swalkinshaw:
Awesome, thanks 👏

PR #1853, opened 2017-03-02
Check for required node version

JulienMelissas:


JulienMelissas:
Right. So this failed because I can't use `semver` before we run `npm install`, and it's gotta do this before it can run an 😂  npm install.

So... I was re-writing the script to not need anything, and it's become a bit intense, but it does work now.

That said, we're talking about using [`engine-strict`](https://docs.npmjs.com/misc/config#engine-strict) even tho it's kinda deprecated. TBD...

cmgriffing:
Hi, I am having a separate issue and came to the issues and PRs to investigate. However I saw this PR and I recently wrote a script like this for our team at work.

While not ideal, you could fetch the semver.js file from github raw. You would then have to eval it, so maybe keeping a checksum as part of the fetch process would be good for security. If you go the checksum route, you will likely need to fetch specifically tagged versions of the file, but it should work.

Just an idea. Cheers.

swalkinshaw:
Closing, we're going with https://docs.npmjs.com/misc/config#engine-strict

PR #1854, opened 2017-03-03
wp_link_pages displayed twice

plaurent75:
wp_link_pages is displayed twice.
Need to add 'echo'=>0 as parameter to avoid it.


retlehs:
thanks!

PR #1855, opened 2017-03-06
Add Aside Post Format & Single Partial

corbinhesse:
Add a single post format (`aside`) to help start themes that utilize this feature. Similar to the `template-custom.blade.php`, only a single partial is available to educate the developer on how to create most post formats if needed.

`single.blade.php` loops through `post_type` then `post_format` if the type is a standard post.

This is simple enough to expand and add more formats if needed and even easier to disable without affecting functionality. Only a single partial file was created.

### DIFF
--- a/src/setup.php
+++ b/src/setup.php
@@ -51,6 +51,12 @@ add_action('after_setup_theme', function () {
     add_theme_support('post-thumbnails');
 
     /**
+     * Enable post formats
+     * @link https://developer.wordpress.org/themes/functionality/post-formats/
+     */
+    add_theme_support('post-formats', ['aside']);
+
+    /**
      * Enable HTML5 markup support
      * @link https://developer.wordpress.org/reference/functions/add_theme_support/#html5
      */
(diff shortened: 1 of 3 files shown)


QWp6t:
We removed post_formats just a few months ago.

https://github.com/roots/sage/pull/1759/commits/78ea970eab0ab449be88e8d3fe4b2194f22d3f83

@retlehs

corbinhesse:
Mind if I ask what for? This just seems like a core part of WordPress' functionality. I've always looked at Sage as a theme that includes all core functionality which allows to easily disable what you do not need.

Menus, widgets, sidebars, post thumbnails, and even title-tags Sage supports. Post formats seems _to me_ to be similar in that regard.

Y'alls theme though. Feel free to close pull request if you do not want formats.

If it is closed, maybe we should do one of the following for consistency:
- Remove the formats if statement on `templates/index.blade/php`
    - Line 14 `@include('partials/content-single-'.get_post_type())`
- Add formats if statement on `templates/single.blade/php`
    - Line 5 `@include('partials/content-single-'.(get_post_type() != 'post' ? get_post_type() : get_post_format()))`

Thanks!

retlehs:
we should remove the formats reference from `index.blade.php`

fwiw it was also removed in underscores, and the [reasoning from the original issue](https://github.com/Automattic/_s/issues/966) aligns with my thoughts:

> Themes approved and set live this way provide a bad user experience as users select one of the post formats and nothing happens

PR #1858, opened 2017-03-14
Add webpack loader for .json and .xml

slackday:
I like adding favicons folder to `assets/images/favicons`. Usually I generate favicons using http://realfavicongenerator.net/ or similar service. Favicon folder contains their own `browserconfig.xml` and `manifest.json`.

This PR keeps those files in `dist/images/favicons/` and makes it easy to just copy the generated `<meta />`-tags to add favicons.

### DIFF
--- a/assets/build/webpack.config.js
+++ b/assets/build/webpack.config.js
@@ -93,6 +93,13 @@ let webpackConfig = {
           name: `vendor/${config.cacheBusting}.[ext]`,
         },
       },
+      {
+        test: /\.(json|xml)$/,
+        include: config.paths.assets,
+        loader: `file?${qs.stringify({
+          name: '[path][name].[ext]',
+        })}`,
+      },
     ],
   },
   resolve: {


slackday:
I'm no webpack expert if anyone see any cons with this approach please let me know!

giacomoalonzi:
@slackday I think is more useful if you use this webpack plugin https://github.com/jantimon/favicons-webpack-plugin. 
It's very simple to use and does everything automatically.

JulienMelissas:
Also, just sayin' - do most people not use [WordPress' built-in favicon support](https://codex.wordpress.org/Creating_a_Favicon#WordPress_Version_4.3_or_later)?

slackday:
@giacomoalonzi Thanks I will check it out!

@JulienMelissas I almost never use the Customizer. The built in support is not good enough with different files for different devices I think.

You can close this PR if you don't want this. But I think I'll stick to my way a bit longer..

JulienMelissas:
@slackday I am also not the biggest customizer fan. Do you know of the differences b/w this implementation and the default WP one (I don't)? I'm not at all interested in closing the PR if there's a good case to be made, and maybe we can explore that favicon webpack plugin? Sorry wasn't trying to shoot your code down at all 👍

mAAdhaTTah:
`json-loader` is built into webpack 2 by default and does something different than what's configured here. I don't know if you want to ship a webpack configuration that conflicts with default expectations.

slackday:
No offence taken :) I didn't mean You personally i meant whoever decides if this is a good idea to merge sorry I write a bit weird and quick sometimes..

Ok good points. This could have unintended side-effects using json-loader.

I will look into favicon-webpack. Maybe it can be configured to be opt in to Sage or we can at least put favicon best practise in Wiki/readme.

JulienMelissas:
So I took a swing at implementing the favicons-webpack-plugin and while it works well, we can't simply use the automatic injecting for the favicon urls, because we're not using the html injector. It would have to make changes to the `head.blade.php` file.

I'm not in favor of merging this pr if we can figure out the json situation, so that people can include custom favicon builds if they want to (like me).

JulienMelissas:
@slackday - I'm following up on this. While we appreciate the code addition, we're going to pass on it for now for 3 reasons:

1. WordPress has built-in favicon support in the customizer, while some of us don't use the customizer, we're going to lean on it for now (lots of people do, and we include some default boilerplate code)
2. I was unable to easily get the favicons-webpack-plugin working easily which was suggested as a replacement
3. Honestly implementing favicons 100% can seriously add tons of files/etc to the build which might be deemed unnecessary as a default feature (I would be in that camp) 

I'd encourage you to add a little guide/how-to in [our discourse](https://discourse.roots.io) and feel free to link back to this code/pr. 😄

PR #1867, opened 2017-04-02
Add Tachyons as a CSS framework option

davekiss:
This framework has been picking up some traction, so thought I'd submit a PR to get it implemented in the Sage 9 generator.

http://tachyons.io/


Log1x:
To add onto this, there a couple other wonderful frameworks out there. Bulma.io being one of them. 

Since we are using postcss as is, maybe we should give the option to use Lostgrid and Rucksack as well. I guess a lot of these in the end would be better suited for the docs though. I'd have no problem at all writing guides on his to get started with postcss, stylus, Vue, etc while using Sage 9.

JulienMelissas:
I'd like to put my vote in to approve this one. Love Tachyons.

Log1x:
+1 for Bulma support

JulienMelissas:
Every single front-end framework/library we add to the Sage repo we need to support minimally. My only issue is when I see Bluma (just as an example) added or requested, it might get to be a bit much. I like Tachyons, but at the same time if everyone wants this that and the other I don't think any of us are seeing the Sage repo be cluttered up. Thoughts?

joaovpmamede:
This is a really great add-on.
Btw @davekiss did you try using [uncss](https://github.com/giakki/uncss) to get rid of unused css that tachyons might add?

retlehs:
shit @davekiss sorry for letting this sit for so long. can you please fix the merge conflicts? then we'll get this in :+1:

davekiss:
@retlehs I believe that should do it, but a quick review would be 👍

PR #1869, opened 2017-04-04
Rename src/ to app/

retlehs:
`src/` (php) had no relation to `dist/` (front-end assets), so this was confusing.

earlier versions of sage 9 used `app`. it was changed to `src` with f752470b337ef6e063055cc99add496f5953e416


Log1x:
Thanks a ton for this.

PR #1871, opened 2017-04-04
Ignore the vendor directory when searching for templates

retlehs:
this isn't applicable until wordpress 4.7.4 lands, see https://core.trac.wordpress.org/ticket/38292


QWp6t:
Also not needed as of #1877

Issue #1870, opened 2017-04-04
Custom templates not loading since latest update


vdrnn:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

## Bug report

It looks like after the latest update of sage, the custom templates aren't working anymore. I tested it on two different installations and for both the custom templates don't appear under page attributes. I guess it has something to do with the move of templates/ to resources/views/ and src/ to app/ but unfortunately I wasn't able to figure out where the problem is. 

Specifically I'm talking about this template/templates `resources/views/template-custom.blade.php`

retlehs:
confirmed bug, thanks for reporting

retlehs:
it looks like wordpress hardcodes looking for page templates beyond 1 level deep 😱, see http://wordpress.stackexchange.com/a/250024 / https://github.com/WordPress/WordPress/blob/503d42fe6bcd2f45b3af78657686f8a27ccb0136/wp-includes/class-wp-theme.php#L1034-L1045

from what i can tell so far, we'd have to duplicate the `get_post_templates` functionality and tell it to search beyond 1 folder deep

...then use the `theme_page_templates` filter to populate the list of template names & filenames

Log1x:
oh jeeze

derkjn:
How about setting the templates in the cache to avoid the issue entirely? Since that method is checking with `$post_templates = $this->cache_get( 'post_templates' );`?

retlehs:
@derkjn how would the issue be avoided by doing that? that solution still requires duplicating a bunch of functionality in the WP theme class

derkjn:
If we preset the list of templates in the cache before wordpress gets to that point, instead of searching for php files in the first level deep, it will return what's in the cache, which basically goes down to accessing the following:
```
function wp_cache_add( $key, $data, $group = '', $expire = 0 ) {
[...]

PR #1876, opened 2017-04-11
Allow custom templates to live in 2 level deep folders

derkjn:
This PR addresses the issue of custom templates not being recognised by WordPress when being placed in folders that are more than 1 level deep as reported in [#1870](https://github.com/roots/sage/issues/1870)


retlehs:
not needed as of #1877

PR #1877, opened 2017-04-14
Move required theme files to sage/resources

QWp6t:
* Resolves issue with WP only looking one-level deep for templates #1870
* get_template_dir() and related functions now point to sage/resources
* Use collect() when assembling views paths


wassim:
@QWp6t As style.css moved to /resources with this PR, I'm getting a `Stylesheet is missing` error.

retlehs:
@wassim you will need to re-activate your theme if you are merging this into an existing install

PR #1878, opened 2017-04-16
fix caption display

drdogbot7:
.wp-caption is the FIGURE element and should extend .figure, contained IMG should extend .figure-img


retlehs:
apologies for the late reply. can you please make this change to add spacing on the part you added?

```diff
- .wp-caption>img{
+ .wp-caption > img {
```

then this is good to go. next sage 8.5.x release probably won't happen until a new version of bootstrap 4 drops

PR #1886, opened 2017-04-25
Remove die/var_dump

schrapel:


swalkinshaw:
😳

Log1x:
Shh.. They'll never know...

PR #1891, opened 2017-05-07
Do not redirect for WP-CLI

schrapel:


QWp6t:
I'd rather check if it's in cli mode (`php_sapi_name() === 'cli'`) than check for WP_CLI specifically. 

And we usually use guards. So final code would be...

```diff
 if ($sage_views !== get_option('stylesheet')) {
     update_option('stylesheet', $sage_views);
+    if (php_sapi_name() === 'cli') {
+        return;
+    }
     wp_redirect($_SERVER['REQUEST_URI']);
     exit();
  }
```

Thoughts?

schrapel:
I'm happy. We're not in a function so this will always have to remain at the end of the file.

rfair404:
Ah, thank you! I've run into the same issue (I assume) that @schrapel has. Attempting to run bitbucket pipelines CI to generate the blade templates since `wp engine` won't allow the blade templates to compile in production env's.

schrapel:
@rfair404 you might want to read my blog https://tobyschrapel.com/2017/05/deploying-sage-9-blade-templates-to-wp-engine-using-gitlab-ci-pipelines/

rfair404:
@schrapel I have and It has helped me a lot, thanks!

PR #1901, opened 2017-06-06
Optimize CSS Assets safe = true

JulienMelissas:
I was running into some issues on production builds with z-indexes as well as a few more advanced background gradients, etc. [This issue](https://github.com/vuejs-templates/webpack/issues/614) linked to a solution I thought I should try which fixed it right away. By default I think this should be included in all builds.

### DIFF
--- a/resources/assets/build/webpack.config.optimize.js
+++ b/resources/assets/build/webpack.config.optimize.js
@@ -14,6 +14,7 @@ module.exports = {
       cssProcessorOptions: {
         discardComments: { removeAll: true },
         autoprefixer: {},
+        safe: true,
       },
       canPrint: true,
     }),


swalkinshaw:
Would be nice to see an example and the size difference.

oxyc:
I've had the issue with Foundation back in Sage 8 as well. Same fix.

JulienMelissas:
> Would be nice to see an example and the size difference.

So I have 2 css files in my build, `main.css` and `critical.css`
Here's with this PR:
```
main: 121kb
critical: 13kb
```

And before:
```
main: 120kb
critical: 13kb
```

This is a production project with 99% of the code written. Not much of a different iirc 😉

JulienMelissas:
Oops, sorry you also said an example. On this one project in `critical.css` I have a z-index set to 2000 for an object that needs to be on top of EVERYTHING on the page at the beginning of the page load (in this example it's an 21+ screen for a brewery). With safe off, it transforms the z-index to 3. It also takes the background gradient animation and totally screws it up. Not exactly sure what it's doing there but I know it doesn't work. Probably a keyframe optimization thing.

oxyc:
Related https://github.com/roots/sage/pull/1610#issuecomment-182969178

JulienMelissas:
@oxyc oh word! looks like this just didn't get ported over in the new Sage.

Issue #1874, opened 2017-04-06
CSS not injected on external URL


nicooprat:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Using external URL on another machine (same network), eg `192.168.1.4:3000`, CSS is not getting injected. 

**What is the expected or desired behavior?**

CSS should be refresh as it is on local URL, eg `localhost:3000`.

---

## Bug report

**Please provide steps to reproduce, including full log output:**

* Install fresh Sage theme, activate, start server. 
* Open one tab with the local URL, another with the external URL given by BrowserSync.
* Make a CSS change, save.

CSS is refreshed on the local tab, not on the external one. 

Notes:

* the BrowserSync notification works.
* No error in console. 
* The local URI **does** work on the same machine though. 
* Tried with another Mac, Windows, IE, iPhone... Same issue everywhere.
* Changes in JS & Blade templates are injected correctly

**Please describe your local environment:**

WordPress version: 4.7.3

OS: Mac OS El Capitan (Chrome latest)

NPM/Node version: 7.4.0

**Where did the bug happen? Development or remote servers?**

Development

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

Didn't find anything.

Thanks in advance!

retlehs:
just tested this out and was unable to reproduce

PR #1902, opened 2017-06-07
Fixes compatibility HMR & Browser Sync with external clients and IE

nicooprat:
Polyfills for IE (tested with IE10+):

* Promise
* EventSource

Fix:

When developing with external clients (`192.168.xx.xx:3000`), HMR could not replace styles because the URL would be replaced by `localhost:3000`, so the `hot-update.json` file returned a 404. Here's a simple workaround, that could certainly be enhanced ([related to this issue](https://github.com/webpack/webpack/issues/443)).

Hope it helps!

PS. Related to this issue https://github.com/roots/sage/issues/1874


QWp6t:
You should be able to make use of `SAGE_DIST_PATH`, as documented here: https://github.com/roots/sage/blob/master/resources/assets/build/config.js#L36-L44

```sh
SAGE_DIST_PATH=192.168.xx.xx:3000/app/themes/sage/dist yarn start
```

Something like that.

I'm closing this for the time being, but feel free to reopen it if you want to try to resubmit a PR that uses the existing solutions we've come up with for this issue (albeit our solutions did not have local development in mind at the time). In the meantime, I'll be looking into this as well.

nicooprat:
Ok thanks, I'll try it. Would it be possible to use the `.npmrc` file to override this kind of configuration? `DEVURL` would be another example, it's pretty useful when working on multisites, but not really handy to write it every time before the command.

QWp6t:
You could edit it in package.json

```json
  "scripts": {
    ...
    "start": "SAGE_DIST_PATH=192.168.xx.xx:3000/app/themes/sage/dist webpack --hide-modules --watch --config resources/assets/build/webpack.config.js",
    ...
  },
```

But that's less than ideal because the IP could change. But that issue isn't alleviated by hard-coding it in .npmrc or anywhere else.

With the next version of Sage, we'll be supporting config-local.json, which will allow you to override config.json settings (the idea being that these are things each particular dev on your team might want to change depending on their development environment, such as proxy URL, dev URL, etc). In which case you could just use config.enabled.watcher to check if you're in watch mode, and if so, then set SAGE_DIST_PATH to the IP address you specify in config-local.json.

In the meantime, you can do all of this in config.js

If you need further help, please post on https://discourse.roots.io/

nicooprat:
Thanks. +1 for local git-ignored file.

For now, anyone interested in storing `DEVURL` in a git-ignored file could just add this line to `.npmrc` at the root of your project: `DEVURL=http://yoursite.dev` and replace this one in `webpack.config.watch.js`: `const target = process.env.npm_config_DEVURL || process.env.DEVURL || config.devUrl;`

PR #1906, opened 2017-06-13
Don't remove comments when clearing files.

### LINKED ISSUE #1905: CSS framework selection needs to no longer fully clear out files
as of #1885 (stylelint addition) we can no longer fully clear out files after someone has selected a CSS framework besides bootstrap

https://github.com/roots/sage/blob/65169986b71a02c73ddd3efe05383783bfbed202/app/lib/Sage/PostCreateProject.php#L43-L48

these files need to still maintain the newly added comments at the top

discourse thread: https://discourse.roots.io/t/yarn-build-failed-because-of-a-stylelint-error/9792

samrap:
Fixes #1905

### DIFF
--- a/app/lib/Sage/PostCreateProject.php
+++ b/app/lib/Sage/PostCreateProject.php
@@ -12,7 +12,7 @@ class PostCreateProject
         $io = $event->getIO();
 
         if ($io->isInteractive()) {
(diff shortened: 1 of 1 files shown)


retlehs:
thank you!

PR #1910, opened 2017-06-21
Controller examples

darrenjacoby:


retlehs:
any thoughts on adding a little bit more? eg. the examples in controller's README

darrenjacoby:
Most of the Controller examples in the README rely on Advanced Custom Fields.  I've followed the JS files with the above, but not sure it's the best use of Controller.  

I feel an App.php and Singular.php (or Single.php and Page.php) might be more obvious that it uses the default WordPress template hierarchy. A good example for Single.php could be getting an array of categories the post belongs to as one can expect that with a default WordPress post. 

Let me know what you think?

retlehs:
gonna merge for now, we can always re-iterate :+1:

PR #1913, opened 2017-06-24
Update filters.php

9585999:
Add browser class to `<body>`


QWp6t:
Thanks for taking the time to submit a PR, but the use case for this seems far too limited to be included in Sage by default. I'll leave this PR open for now so that you or someone else can try to make a case for it, as it's quite possible that I'm completely overlooking its usefulness.

oxyc:
Also doesn't work with proxy/page caching which could easily cause problems down the road if used.

JulienMelissas:
Yeah, I think I'd pass on this one as well.

PR #1914, opened 2017-06-25
Add WooCommerce support for Controller

darrenjacoby:
Passing the data to woocommerce allows Controller to use the data in WooCommerce template files.

### DIFF
--- a/app/filters.php
+++ b/app/filters.php
@@ -56,6 +56,9 @@ collect([
  * Render page using Blade
  */
 add_filter('template_include', function ($template) {
+    if (is_woocommerce()) {
+        $template = 'woocommerce';
+    }
     $data = collect(get_body_class())->reduce(function ($data, $class) use ($template) {
         return apply_filters("sage/template/{$class}/data", $data, $template);
     }, []);
--- a/app/filters.php
+++ b/app/filters.php
@@ -56,13 +56,14 @@ collect([
  * Render page using Blade
  */
 add_filter('template_include', function ($template) {
-    if (is_woocommerce()) {
-        $template = 'woocommerce';
-    }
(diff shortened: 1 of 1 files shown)


darrenjacoby:
App.php passes the variables to all pages, and not be overridden using the Controller hierarchy, which may not be ideal.  The above is checking if it's a WooCommerce template, and then adding those to woocommerce.blade.php which gets the WooCommerce template from woocommerce/.  It also allows users to have a SingleProduct.php Controller file for example. 

I haven't looked into `woocommerce_template_loader_files`, am I missing something there? I've based the above solution on this issue, https://github.com/soberwp/controller/issues/16

QWp6t:
Yeah, so far in all my tests, I've been able to get this to work without any issues. I'm not really sure which method you guys are using to make your theme support WooCommerce, but I used a few different techniques. The easiest to hook into `woocommerce_template_loader_files` and tell it to look for `woocommerce.blade.php`, but if you want the increased granularity of single-product, archive-product, etc., then you can hook into `woocommerce_locate_template` and return `template_path($template)`. (You can do the same with `wc_get_template_part` to include WooCommerce partials.)

The former will make controller/Woocommerce.php available while the latter will make controller/SingleProduct.php available.

For what it's worth, WooCommerce recommends making use of template hooks instead of overwriting their template files. But I will still document all of this before Sage 9 launches.

darrenjacoby:
@QWp6t thanks!  That information is def all helpful!

mtx-z:
I found a way to pass correctly all variables to Woocommerce Views, don't know if it's the right way but it works as for now : https://discourse.roots.io/t/controller-and-woocommerce/9469/5

Issue #1915, opened 2017-06-27
[feature] BrowserSync disable opening in a new window (extend the configuration)


jmarceli:
It might be only my personal preference but I think that the user should be able to select if they want to open new browser window after dev server initialization (`npm start`) or generally set any BrowserSync options in config.

Here is what needs to be done:

- add `browserSync` key to the `resources/assets/config.json` file
- adjust `resources/assets/build/webpack.config.watch.js` by adding `advanced` key to the **BrowserSyncPlugin** initialization options
- optionally add the `browserSync` key default values inside `resources/assets/build/config.js`

One more important thing. The build is using [browsersync-webpack-plugin](https://github.com/QWp6t/browsersync-webpack-plugin) **not** the [browser-sync-webpack-plugin](https://github.com/Va1/browser-sync-webpack-plugin) which seems to be much more popular. I think it should be mentioned somewhere to ease searching (options are slightly different for both plugins).

If you think that adding BrowserSync options to the `config.json` is a good idea I would be happy to create a PR.

JulienMelissas:
Sage users aren't just guys so I fixed that language for ya 😉 

I'm totally open to a PR! The browser window setting has been one I've thought about before, and in general I don't personally mind if config.json has a little bit more. The browsersync plugin we're using has a bit more stuff specific to the sage build, so we'll def. want to keep it - where are you suggesting this mention is? I haven't heard of anyone  having an issue of someone confusing them so far, but maybe it's been a thing.

QWp6t:
Done: https://github.com/roots/sage/pull/1919/commits/a429115fb26e3d67c41950c1fbffb6fb9a451321

Also, there's going to be support for a config-local.json. So each developer on the team can override any of the config settings as needed.

PR #1916, opened 2017-06-28
Add browserSync key to config.json

jmarceli:
As described in #1915 I've added a browserSync key to the `resource/assets/config.json` where any [BrowserSync related options](https://browsersync.io/docs/options/) can be set.


jmarceli:
Fixed CI bugs in PR #1918.

PR #1918, opened 2017-07-01
Add browserSync key to config.json

jmarceli:
As described in #1915 I've added a browserSync key to the resource/assets/config.json where any BrowserSync related options can be set.

### DIFF
--- a/resources/assets/build/config.js
+++ b/resources/assets/build/config.js
@@ -25,6 +25,9 @@ const config = merge({
     watcher: !!argv.watch,
   },
   watch: [],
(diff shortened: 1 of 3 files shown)


QWp6t:
Went ahead and added an `open` option to the browsersync plugin. I pushed a commit to sage that makes use of that. I already marked you as the author so you still get credit:

https://github.com/roots/sage/pull/1919/commits/a429115fb26e3d67c41950c1fbffb6fb9a451321

jmarceli:
OK. Thanks for the info.

PR #1923, opened 2017-07-09
WooCommerce support

QWp6t:
I'm personally against adding this to Sage out of the box. I think it would be better to  put it on our docs, and that's probably what will wind up happening, but I figured I would go ahead and submit a PR anyway for now if for no other reason than to serve as an interim example for using Blade with third-party plugins.

### DIFF
--- a/app/setup.php
+++ b/app/setup.php
@@ -30,6 +30,11 @@ add_action('after_setup_theme', function () {
     add_theme_support('soil-relative-urls');
 
     /**
+     * Declare WooCommerce support
+     */
+    add_theme_support('woocommerce');
+
+    /**
      * Enable plugins to manage the document title
      * @link https://developer.wordpress.org/reference/functions/add_theme_support/#title-tag
      */
--- a/app/woocommerce.php
+++ b/app/woocommerce.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace App;
+
+/**
+ * WooCommerce Support
+ */
+add_filter('woocommerce_template_loader_files', function ($search_files, $default_file) {
+    return filter_templates(array_merge($search_files, [$default_file, 'woocommerce']));
+}, 100, 2);
+add_filter('woocommerce_locate_template', function ($template, $template_name, $template_path) {
+    $theme_template = locate_template("{$template_path}{$template_name}");
+    return $theme_template ? template_path($theme_template) : $template;
+}, 100, 3);
+add_filter('wc_get_template_part', function ($template, $slug, $name) {
+    $theme_template = locate_template(["woocommerce/{$slug}-{$name}", "woocommerce/${name}"]);
+    return $theme_template ? template_path($theme_template) : $template;
+}, 100, 3);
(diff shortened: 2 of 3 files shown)


christianmagill:
Maybe a composer option?

arthurkirkosa:
I keep running into `Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function startSection() on null in ... cache ... line 1` with this approach

This is the content of that file.
```php
<?php $__env->startSection('content'); ?>
  <?php while(have_posts()): ?> <?php (the_post()); ?>
    <?php echo $__env->make('partials.page-header', array_except(get_defined_vars(), array('__data', '__path')))->render(); ?>
    <?php echo $__env->make('partials.content-page', array_except(get_defined_vars(), array('__data', '__path')))->render(); ?>
  <?php endwhile; ?>
<?php $__env->stopSection(); ?>

<?php echo $__env->make('layouts.app', array_except(get_defined_vars(), array('__data', '__path')))->render(); ?>
```

I've created `resources/views/woocommerce/content-product.blade.php` that has the content from `page.blade.php`

jloproductions:
> I keep running into Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function startSection() on null in ... cache ... line 1 with this approach

Did you @arthurkirkosa find any solutions for this? I'm getting the same error.

arthurkirkosa:
```php
// resources/views/woocommerce.blade.php
@extends ('layouts.app')

@section ('content')
@php (woocommerce_content())
@stop
```

```php
// resources/views/woocommerce/archive-product.php
<?php echo App\Template('woocommerce');
```

In a partial I have
```php
<div class="row mb-5">
    @while ($products->have_posts()) @php ($products->the_post())
    <div class="col-sm-4">
        <div class="product product-border">
            <a class="woocommerce-loop-product__link" href="{{ get_permalink() }}">
                {!! woocommerce_get_product_thumbnail() !!}
                <div class="woocommerce-loop-product__title">{{ the_title() }}</div>
            </a>

            {!! woocommerce_template_loop_price() !!}
        </div>
    </div>
    @endwhile
</div>
```

```php
<?php

namespace App\Traits;

use WP_Query;

trait Products
{
    public function products()
    {
        return new WP_Query([
            'post_type' => 'product',
            'posts_per_page' => 3,
            'tax_query' => [
                [
                    'taxonomy' => 'product_visibility',
                    'field' => 'name',
                    'terms' => 'featured',
                ],
            ],
        ]);
    }
}
```

Not exactly what I wanted... but it's the best I got so far... would have liked to be able to override the actual HTML structure, but... ¯\_(ツ)_/¯

Maybe someone can help us with that part.

I was able in Sage 8 to do it... but not sure how to in Sage 9 (and I'm a Laravel developer :D)

nboliver:
@QWp6t The pr you posted was really helpful. Is it possible to use blade templates instead of php templates though?

PR #1924, opened 2017-07-18
Switch to prettier?

QWp6t:
It seems to have taken off in the js world.

It would alleviate issues with users running into eslint/stylelint errors.

This PR also adds a git hook to run prettier, which ensures that future commits are consistently styled.


samrap:
> By far the biggest reason for adopting Prettier is to stop all the on-going debates over styles. It is generally accepted that having a common style guide is valuable for a project and team but getting there is a very painful and unrewarding process.

If I'm understanding it right, Prettier decides the code style for you. This is a really bad direction to take as the agency I work for has a specific set of style rules we apply across all our projects (Sage + WP, Laravel, Magento, etc). I imagine it would be the same thing for many other users. 

We would be unable to use Sage as-is if Prettier is merged into the theme. Forking and updating each time there is an update would cause a big headache. 

I think we need to allow the developers to choose their coding styles as I am sure many others like us have spent significant time setting specific standards. For those who have not, the existing rules from eslint and stylelint should suffice. 

> Prettier does nothing to help with [code quality] rules. They are also the most important ones provided by linters as they are likely to catch real bugs with your code!

Regardless of using Prettier, eslint should still be included for code quality rules.

---

Overall if I had a say it would be to vote against this PR. Thanks for all your hard work on Sage, for reading this, and for working to continually improve this awesome project!

QWp6t:
Yeah, I overestimated what `prettier` was capable of.

Since it completely rewrites all of the code it analyzes, I assumed that it would address code quality as well. I should have probably looked into it a bit more before deciding to go all in with it.

PR #1942, opened 2017-08-11
update bootstrap v4 beta1

kukac7:


retlehs:
thanks for the PR! when updating `package.json` deps, please also commit the changed `package-lock.json` and `yarn.lock` files

MWDelaney:
It seems that Bootstrap 4 no longer uses the `$brand-primary` naming scheme for colors. This dependency update will break Sage initially until `_variables.scss` is updated accordingly.

retlehs:
@MWDelaney nice, thanks for pointing that out!

Issue #1945, opened 2017-08-12
Popper not working in Beta 4


MWDelaney:
## Submit a feature request or bug report

- [X] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [X] This is a bug report
- [X] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [X] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Popper (required by Bootstrap 4 Beta for dropdowns) is not working in Sage9 Beta 4. I banged my head against an existing theme I'd ported into Beta 4 for a while with no luck, and eventually tried a plain vanilla Beta 4 and found the same results. 

Upon clicking a dropdown trigger, I get the following message:
```
TypeError: Object is not a constructor (evaluating 'new Popper(element, this._menu, this._getPopperConfig())')
```

I tried importing Popper in `main.js` and nothing changed.

**What is the expected or desired behavior?**

Popper should work, as it's required for Bootstrap 4 functionality.

---

## Bug report

**Please provide steps to reproduce, including full log output:**

1. Install a fresh copy of Sage9 Beta 4 with composer, run basic setup and `yarn` steps
1. Add the following Bootstrap 4 example code to a view:
````
<div class="dropdown">
  <button class="btn btn-secondary dropdown-toggle" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
    Dropdown button
  </button>
  <div class="dropdown-menu" aria-labelledby="dropdownMenuButton">
    <a class="dropdown-item" href="#">Action</a>
    <a class="dropdown-item" href="#">Another action</a>
    <a class="dropdown-item" href="#">Something else here</a>
  </div>
</div>
````
3. Rerun `yarn build` and see the console errors when you try to trigger the dropdown.

**Please describe your local environment:**

WordPress version: 4.8.1

OS: MacOS X

NPM/Node version: v7.4.0

**Where did the bug happen? Development or remote servers?**

Development

[...]

PR #1946, opened 2017-08-12
Fix Popper for Bootstrap 4

MWDelaney:
This is in reference to #1945, I figured out what was wrong. Bootstrap's [dropdown documentation](https://getbootstrap.com/docs/4.0/components/dropdowns/) indicates the "umd" version of Popper (poorly; you have to look at the actual URL they link to) which fixes the issue of dropdowns not working in Sage Beta 4.


QWp6t:
Thanks!

PR #1948, opened 2017-08-13
Change default Controller path to app/controllers

darrenjacoby:
Move the Controller default path to app/controllers.

Controller PR to change which I will merge if/when this is merged—https://github.com/soberwp/controller/pull/40


retlehs:
can you pls rebase since i just merged #1912? then we could prolly coordinate the merges so we don't screw up any new builds

darrenjacoby:
Cool, will rebase this weekend.

PR #1951, opened 2017-08-17
Issue #1950: Fix lint:styles script not matching files in deep subdirectories

oxyc:
Fix #1950 by not relying on the user's shell but rather passing the unexpanded pattern directly to `stylelint` which in turn expands it using `node-glob`.

Issue #1950, opened 2017-08-17
Glob pattern in lint:styles is not expanded in all shells


oxyc:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Currently running `npm run lint:styles` does not lint files in 2-level deep directories.

**What is the expected or desired behavior?**

The glob pattern `/**/*.{scss,..}` should be expanded to match all files in any depth.

---

## Bug report

**Please provide steps to reproduce, including full log output:**

1. Ensure your PATH's `sh` bin points to a) bourne shell b) bash < 4 c) bash 4+ with `extglob` disabled.
2. `mkdir resources/assets/styles/components/foo`
3. `echo "foo" > resources/assets/styles/components/foo/bar.scss`
4. `yarn run lint:styles`

```
yarn run v0.24.5
$ stylelint resources/assets/styles/**/*.{css,sass,scss,sss,less}
✨  Done in 1.85s.
```

While as running the command manually from Bash 4:

```
./node_modules/.bin/stylelint resources/assets/styles/**/*.{css,sass,scss,sss,less}

resources/assets/styles/components/foo/bar.scss
 1:1  ✖  Unknown word   CssSyntaxError
```

**Please describe your local environment:**

OS: macOS Sierra
NPM/Node version: v8.1.0

**Where did the bug happen? Development or remote servers?**

Development

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

Not that I'm aware of.

oxyc:
Related https://discourse.roots.io/t/yarn-build-error-exit-code-2/10225

PR #1955, opened 2017-08-20
Followup #1950: Fix lint:styles task with cmd.exe

oxyc:
See https://github.com/roots/sage/pull/1951#issuecomment-323593159

Note that I haven't tested this on windows.


retlehs:
can someone please confirm this works?

oxyc:
I think it's fairly safe to merge as @mtx-z confirmed it fixed his issue and it's also what `eslint` now suggests with a note about this exact issue https://github.com/eslint/eslint/commit/6f88546572b6ee8bc93a151f1df2b4f85c2abafe

PR #1958, opened 2017-08-22
Update version number in composer install command.

MikeiLL:
Simple doc commit.


retlehs:
well shit. it's too late now since 8.5.3 is already tagged. updated the base target to `sage-8` on this PR. thanks for catching this

MikeiLL:
I thought about adding something like `# match current version` after the command.

PR #1919, opened 2017-07-02
Sage 9.0.0-beta.4?

QWp6t:
I started off just wanting to update some dependencies, but I wound up getting a little carried away. lol

Gonna work on squashing some bugs now.

To see the new post-create-project routine, run the following:

```sh
composer create-project --prefer-dist --stability=dev roots/sage:dev-update-dependencies YOUR-FOLDER-NAME-HERE
```

PR #1960, opened 2017-08-22
Move /controllers to app, remove app/lib/Sage in readme

greatislander:
Aligns readme.md with the relocation of `controllers` in #1954 and #1959, and reflects the removal of `app/lib/Sage` in #1919.


retlehs:
thanks!

Issue #1932, opened 2017-07-28
[Feature] Directives configuration file


kaisermann:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Currently there's no convenient way and place to put custom directives (the `@asset` is created at the end of the `app/setup.php`)

**What is the expected or desired behavior?**

It would be good to have something like `config/directives.php`. [See Example](https://github.com/wilsonpinto/laravel-blade-directives#installation)

---

## Feature Request

**Please provide use cases for changing the current behavior:**

It's easier and clearer to define directives in a dedicated file. 

Part of my current `directives.php`:

```blade
<?php

$getBladeArgs = function( $expression ) {
	return explode(', ', str_replace(['(', ')'], '', $expression));
};
$fnEndWhile = function () { return '<?php endwhile; wp_reset_query(); ?>'; };

return [
	/*
	|--------------------------------------------------------------------------
	| Directives
	|--------------------------------------------------------------------------
	|
	| Custom directives
	|
	*/

	// Create @asset() Blade directive
	'asset' => function ( $asset ) {
		return "<?php echo " . __NAMESPACE__ . "\\asset_path({$asset}); ?>";
	},

	// Creates @mainquery Blade directive
	 'mainquery' => function () {
		return '<?php while(have_posts()) : the_post(); ?>';
	},

	// Creates @mainquery Blade directive
	'endmainquery' => $fnEndWhile,

	// Creates @customquery(\WP_Query $queryObj) Blade directive
	'customquery' => function( $queryObj ) {
		$output = "<?php while ({$queryObj}->have_posts()) : ?>";
		$output .= "<?php {$queryObj}->the_post(); ?>";
		return $output;
	},

	// Creates @endcustomquery Blade directive
	'endcustomquery' => $fnEndWhile,

[...]

PR #1962, opened 2017-08-30
Bump illuminate/support to 5.5

Log1x:
Some new goodies with the release of Laravel 5.5

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
   "require": {
     "php": ">=7",
     "composer/installers": "~1.0",
-    "illuminate/support": "~5.4",
+    "illuminate/support": "~5.5",
     "roots/sage-lib": "~9.0.0-beta.4",
     "soberwp/controller": "~9.0.0-beta.4"
   },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "b58fbe0c86d06990c1c7ef7414df19e6",
+    "content-hash": "06839357ce2c107825c336e0c746e37a",
     "packages": [
         {
             "name": "brain/hierarchy",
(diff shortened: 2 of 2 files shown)


Log1x:
[`@switch`](https://laravel.com/docs/5.5/blade#switch-statements) seems to work as intended as well as everything else stock but I seem to get an error with trying to use the new [Custom If Statements](https://laravel.com/docs/5.5/blade#custom-if-statements).

**Example Code**
```
$sage = sage('blade')->compiler();

/**
 * Create @test() Blade directive
 */
$sage->if('test', function ($field) {
    return true;
});
```

```
@test 
   Hello
@endtest
```

results in:

```
Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function getEngineResolver() on null in /mnt/d/Development/web/sites/flux/web/app/themes/flux/vendor/illuminate/support/Facades/Blade.php on line 17
```

patram1121:
The bump to 5.5 cased issues for my current project.  In the following blade template, the last @php (right before $link) did not convert to <?php in the blade cache.

Note sure if this is a 5.5 issue or there is a change I need to make with my code to make it compatible.

```
    @if (have_rows('acf_repeater))
      @php
        $count = 1
      @endphp
      <div class="col-sm-3">
          @while (have_rows('acf_repeater')) @php(the_row())
            @php
              $link = sanitize_title($title);
            @endphp
```

Was able to workaround by forcing 5.4 in composer:

```
"illuminate/support": "~5.4.0",
```

Log1x:
Hmm. I did some testing and come across the same issue. It appears to be a problem with the shorthand PHP directive. I don't really understand what is causing it though. In some cases, it works just fine, but in others, not.

webstractions:
I think you need a closing `@php` on the previous line:

```
@while (have_rows('acf_repeater')) @php(the_row()) @endphp
```
And FWIW, using this blade feature frequently may be a signal that you have too much logic embedded within your template. See [Blade Docs](https://laravel.com/docs/5.5/blade#php)

webstractions:
@Log1x I am pretty sure there isn't an $app container in Sage. Thus the error. The `if()` method is running thru the Blade Facade and is trying to access `$app['view']` which doesn't exist.

```
   /**
     * Get the registered name of the component.
     *
     * @return string
     */
    protected static function getFacadeAccessor()
    {
        return static::$app['view']->getEngineResolver()->resolve('blade')->getCompiler();
    }
```

webstractions:
Just dawned on me about the other Sage components, `sage-installer` and `sage-lib`, which also need to be bumped up to 5.5 as well. They are at ~5.4 for Illuminate dependencies for Console, Filesystem, Config, and wait for it, View.

Those two packages need to be updated for this PR to play nice.

Log1x:
I'll give that a shot and do some testing and if all is well I'll do a PR over that way too so we can hopefully get this bumped.

webstractions:
Sounds like a plan. Pretty sure that won't solve the custom if() statement though. You and @QWp6t talked about the App Container in #1932, which would fix this.

webstractions:
@Log1x I have a 5.5 install running with no problems so far. Upgrading across the board (lib, installer, sage) doesn't have any ill effects.

The `if()` Blade directive is a sticky point. I am getting the same error that you did. In order to get around that, you have to install the complete Laravel package and create an App container. That would be going to far for what Sage is meant to be.

Coincidentally, this is not the first time something like this has come up (ie: cherry-picking Illuminate components). Take a look at Matt Stauffer's [Torch](https://github.com/mattstauffer/Torch). They have been trying to get Artisan support to work without downloading the entire package as well. 

This just won't work without Illuminate\Foundation, which doesn't have composer.json to install from. And weirdly, Laravel Lumen has it's own version of Artisan piecemealed together and is kind of flakey. Seems they couldn't get around that either.

webstractions:
@Log1x I found a workaround for getting the Blade `if()` method to work!

In `setup.php` add the following to the Sage Setup options where the `singletons` for assets and Blade are being bound to the container. 

```php
    // Set the Facade application container to the Sage Container
    \Illuminate\Support\Facades\Facade::setFacadeApplication(sage());
```
This not only fixes the `if()` method, but should allow us to use facades as well. The only thing missing is the AliasLoader, which is part of Illuminate\Foundation.

PR #1963, opened 2017-08-31
Update config.json

9585999:
change watch path controllers

### DIFF
--- a/resources/assets/config.json
+++ b/resources/assets/config.json
@@ -14,8 +14,8 @@
   "cacheBusting": "[name]_[hash:8]",
   "watch": [
     "app/**/*.php",
+    "app/controllers/**/*.php",
     "config/**/*.php",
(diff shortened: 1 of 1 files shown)


QWp6t:
That line should be deleted entirely

resources/assets/config.json
```diff
@@ -14,8 +14,8 @@
   "cacheBusting": "[name]_[hash:8]",
   "watch": [
     "app/**/*.php",
+    "app/controllers/**/*.php",
```
kalenjohnson: No need for this, the line above has everything covered

9585999: did not dare to delete)

kalenjohnson:
Thanks!

9585999:
@kalenjohnson no problem :)

PR #1986, opened 2017-10-30
Update controller examples

retlehs:
spurred from https://discourse.roots.io/t/best-practice-resources-for-blade/8341/141?u=benword

regarding the `About.php` file prior to this PR... it did nothing. ideally you'd create a `template-about.php` controller _and_ template file, then drop this in the controller:

```
<?php

namespace App;
use Sober\Controller\Controller;

class TemplateAbout extends Controller
{

}
```


robrecord:
Thank you for this, I was mightily confused!

PR #1991, opened 2017-11-10
use relative link for publicPath

seahken:
using '/wp-content/themes/sage' will cause relative link issue when compiling assets if wordpress is installed in subdirectories. Changed to relative path './../..' ensures working on all installation.


QWp6t:
I don't think the readme needs to be changed. It says that it "should reflect your WordPress folder structure" and then we give an example of how it would be set in typical environments.

If you need the public path to be set dynamically, we support that as well. https://github.com/roots/sage/blob/master/resources/assets/build/config.js#L43-L60

PR #1992, opened 2017-11-13
Adapt to use Bootstrap 4 Beta 2 functions and variables

MWDelaney:

PR #1993, opened 2017-11-13
Move variables and Bootstrap lines to autoload

MWDelaney:
See discussion on #1992

### DIFF
--- a/resources/assets/styles/main.scss
+++ b/resources/assets/styles/main.scss
@@ -1,3 +1,5 @@
+@import "common/variables";
+
 /** Import everything from autoload */
 @import "./autoload/**/*";
 
--- a/resources/assets/styles/main.scss
+++ b/resources/assets/styles/main.scss
@@ -1,7 +1,3 @@
-@import "~bootstrap/scss/functions";
-@import "common/variables";
-@import "~bootstrap/scss/variables";
-
 /** Import everything from autoload */
 @import "./autoload/**/*";
 
--- a/resources/assets/styles/autoload/_bootstrap.scss
+++ b/resources/assets/styles/autoload/_bootstrap.scss
@@ -1,3 +1,1 @@
-@import "~bootstrap/scss/functions";
-@import "../common/variables";
 @import "~bootstrap/scss/bootstrap";
--- a/resources/assets/styles/autoload/_bootstrap.scss
+++ b/resources/assets/styles/autoload/_bootstrap.scss
@@ -1,1 +1,3 @@
+@import "~bootstrap/scss/functions";
+@import "../common/variables";
 @import "~bootstrap/scss/bootstrap";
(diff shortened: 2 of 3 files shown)


resources/assets/styles/main.scss
```diff
@@ -1,7 +1,3 @@
-@import "~bootstrap/scss/functions";
-@import "common/variables";
```
LeoColomb: Actually, this line can be kept in the `main.scss` file.
If we remove it, other frameworks variables are not loaded, and `~bootstrap/scss/functions` doesn't seem to override variables.

retlehs:
need to address `@import "common/variables";` being removed from `main.scss`

we could add the variables import to the top of all the [other autoload files in sage-installer](https://github.com/roots/sage-installer/tree/master/src/Presets/stubs), but that doesn't work for the 'none' option

..although we could add the autoloader w/ the variables import to the none option ¯\_(ツ)_/¯

any ideas?

resources/assets/styles/main.scss
retlehs: i think the concern is using a bootstrap function in your variables

MWDelaney: Bootstrap themselves [use functions in variables](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss) so it's confusing if we can't do the same in Sage.

LeoColomb: You can always take the example from zurb/foundation, calling `functions.scss` inside `variables.scss`:

https://github.com/zurb/foundation-sites/blob/eaa79c89a8c5dc89f8fa1e6916ce5dea6cc55866/scss/settings/_settings.scss#L63

MWDelaney: You mean [here](https://github.com/roots/sage-installer/blob/master/src/Presets/stubs/Bootstrap/styles/common/_variables.scss) in sage-installer?

That would work. Is that less messy than including `common/variables.scss` in each autoload? Is it more or less flexible in the future?

LeoColomb: Yes, yes, yes. 😆 👍 
At least, it should be, imo.

MWDelaney: I'm comfortable with that solution. @retlehs do you have an opinion on including the BS functions right in the variables.scss file?

And forgive my ignorance, should I do that on this PR or is that a sage-installer only thing?

retlehs: sounds good to me, you'll want to do that both on this repo (as it should function properly w/ bootstrap without the installer) and sage-installer

MWDelaney:
I guess nobody can say we didn't thoroughly discuss the solution this time! I'll take a stab at sage-installer tomorrow unless somebody else gets there first <--hint ;)

retlehs:
updated sage-installer + new version tagged, should be all good now. thanks @MWDelaney & @LeoColomb!

PR #2008, opened 2017-12-18
Allow `no-console` development

alwaysblank:
So far as I can tell, .eslint files don't allow for internal conditionals, so this PR implements a "production" eslint file that is used when the production flag is set. Otherwise, it uses `package.json`. I structured it this way so that text editors will hopefully not be confused about which eslint file to read rules from. Rules can also be directly passed in the webpack config, but I felt that this would scale poorly and a separate file makes things a bit easier to understand. This change would also allow for easy customization for other dev vs prod eslint rules.

This would resolve issue #2006.

### DIFF
--- a/resources/assets/build/config.js
+++ b/resources/assets/build/config.js
@@ -20,7 +20,7 @@ const config = merge({
     root: rootPath,
     assets: path.join(rootPath, 'resources/assets'),
     dist: path.join(rootPath, 'dist'),
-    eslintProd: path.join(rootPath, '.eslintrcProd'),
+    eslintProd: path.join(rootPath, '.eslintrc.production.json'),
   },
   enabled: {
     sourceMaps: !isProduction,
--- a/resources/assets/build/config.js
+++ b/resources/assets/build/config.js
@@ -20,7 +20,6 @@ const config = merge({
     root: rootPath,
     assets: path.join(rootPath, 'resources/assets'),
     dist: path.join(rootPath, 'dist'),
-    eslintProd: path.join(rootPath, '.eslintrc.production.json'),
   },
   enabled: {
     sourceMaps: !isProduction,
--- a/resources/assets/build/config.js
+++ b/resources/assets/build/config.js
@@ -20,6 +20,7 @@ const config = merge({
     root: rootPath,
     assets: path.join(rootPath, 'resources/assets'),
     dist: path.join(rootPath, 'dist'),
+    eslintProd: path.join(rootPath, '.eslintrcProd'),
   },
   enabled: {
     sourceMaps: !isProduction,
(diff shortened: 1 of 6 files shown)


swalkinshaw:
`.eslintrcProd` should probably be `.eslintrc.production.json`. ref: https://eslint.org/docs/user-guide/configuring#configuration-file-formats

mAAdhaTTah:
If you make it `.eslintrc.js`, I feel like you should be able to add conditionals, but I haven't done it myself.

swalkinshaw:
Yeah I did wonder that if we're adding an actual file, then maybe all the others should be moved from `package.json` in they're own file too. So we either use conditionals, or two files.

alwaysblank:
IMO I'd lead toward two files (i.e. `.eslintrc` and `.eslintrc.production.json`): It seems cleaner and easier to understand, and makes it very clear where rules go.

LeoColomb:
Tt seems more common to have a single `.eslintrc.js`.
(See [some](https://github.com/airyland/vux/blob/bd46fb824af4fe7fdadd7832be7174c17b6b0a35/.eslintrc.js) [popular](https://github.com/nuxt/nuxt.js/blob/ae754af3625a18a698591a0444e257f7bc6bb635/.eslintrc.js) [examples](https://github.com/vuetifyjs/vuetify/blob/646a48a2902eccc5a9cea041794e45252e0bddb9/.eslintrc.js))

mAAdhaTTah:
I think taking that logic to the extreme would make things more confusing, especially cuz if you wanted to add the rule to both, you now have to add it in two places, which is going to be the more likely update. Generally speaking, throughout Sage, you've single single files with conditionals based on the env (e.g. the webpack build [has this](https://github.com/roots/sage/blob/master/resources/assets/build/webpack.config.js#L188-L190), and I think that's more common and easier to understand.

As a comparison, I just started a project with Create React App and ejected, and found two separate webpack configurations, and I kinda hate it. Once I start tinkering around in there, I feel like I'm doing everything twice (not to mention it feels kind of over engineered for my purposes, but that's a separate issue).

swalkinshaw:
One of Sage's goals is to keep dev and prod as close as possible. Using a single file would probably encourage this. Right now there'd only be a single difference (debugger related things).

alwaysblank:
Solid arguments, my bad for not digging a bit deeper initially. Setting conditionals inside `.eslintrc.js` is perfectly easy, and VS Code picks them up just fine. This update to the PR implements the solution that seems to be the consensus.

resources/assets/build/config.js
```diff
@@ -20,6 +20,7 @@ const config = merge({
     root: rootPath,
     assets: path.join(rootPath, 'resources/assets'),
     dist: path.join(rootPath, 'dist'),
+    eslintProd: path.join(rootPath, '.eslintrc.production.json'),
```
swalkinshaw: 🔥

Issue #1868, opened 2017-04-02
Laravel Mix


Log1x:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

## Feature Request

There has been a good amount of chatter behind the scenes on if Sage 9 should move to Laravel Mix. 

I've heard it has its fair share of issues but most seem to be getting patched as they arise. Jeffery Way who built Laravel Mix has an incredible reputation and Mix once all the remaining bugs are kinked out will be quite awesome.

It allows the use of stylus, sass, less, vue, and more out of the box with ridiculously easy configuration.

I'd love some community and developer feedback on this. If it gains traction but none of the developers have the time to work on it, I'd be willing to work on it in a PR as well as update the Sage 9 docs. 

Main things to take into consideration is doing an initial mix configuration that allows us to remap all the directories, and extend various plugins/modules like `purifyCss` to work in Sage's environment. 

And of course, customizing the notification icon to show a roots icon instead. :)

samrap:
Laravel Mix is very new, and Sage 9 is already in Beta. The beta phase is generally for bug fixes after all features have been implemented. To rewrite Sage 9 to be built off of Mix at this point in the game would not make sense IMO, especially when there is nothing wrong with Sage 9 as it stands. What does Mix have to offer that Sage doesn't that is worth taking Sage 9 development back to Alpha? To me a fluent API for people who don't understand webpack is not worth killing what the Sage developers have worked so hard for, especially when it works beautifully. Many people, myself and company included, are anticipating the release of Sage 9 in the near future. Porting everything over to Mix just doesn't make sense to me.

While I'm a huge Laravel fan I do not think that it makes sense to drastically change Sage 9 for the sake of a shiny new package when it is so close to being ready.

Log1x:
[...]

PR #2011, opened 2017-12-24
Switch to laravel-mix

QWp6t:
Ideally I'd like to check these boxes before merging. 

The following still needs to be fixed/added...
- [ ] support for css injection (it injects, but then reloads anyway)
- [ ] support for js injection + hmr (not super important, tbh)
- [ ] vendor extractions for supported presets (bootstrap, foundation, etc)
- [ ] restore support for user-defined cache-busting naming convention (e.g., `[name]_[hash:8]`)

The following still needs to be tested (these should be simple to test, i just didn't do it yet)...
- [ ] watch mode: adding background image (does image copy to dist?)
- [ ] watch mode: removing background image (does image get removed from dist?)
- [ ] copied images that aren't referenced in css/js should get cache-busted
- [ ] source maps


retlehs:
from 36 direct dev dependencies to 11 :scream: 

x-ref https://github.com/roots/sage-installer/issues/1 & https://github.com/roots/sage/issues/1868

QWp6t:
Just to keep everyone in the loop, we're hesitant to merge this now because preliminary benchmarks are showing that laravel-mix is _significantly_ slower at compiling compared to our current setup. I'm not sure exactly what's different. I'm looking into it. But we're talking about differences of 10-20s, far too substantial to ignore.

dmgawel:
@QWp6t [CSS URL processing in Mix](https://github.com/JeffreyWay/laravel-mix/blob/master/docs/css-preprocessors.md#css-url-rewriting) is known for its slow performance in some cases. Did you tried with this option disabled? https://github.com/JeffreyWay/laravel-mix/issues/49#issuecomment-284256971

retlehs:
Yes, that didn’t help

myleshyson:
This looks awesome so far! Just out of curiosity though, are the benchmarks based off just the initial install of sage or are you using a sample project? When I pulled in the laravel-mix branch I found that there isn't much difference in compiling time, but then again I'm just using the base sage theme. 

It seemed like currently there were two seperate scripts running with the laravel-mix branch: npm lint -s and then the build script. When I removed that pre-build script and replaced it by adding eslint and styleint to the webpack configuration, it seemed to start compiling faster.

Here's the compiling times on my laptop.

Build Script | Laravel Mix | Sage Setup
------------|-------------|------------
build | 6793ms | 1623ms
build:production | 18734ms | 18354ms
start (initial) | 4413ms | 6578ms
start (small css change) |2749ms | 3207ms

The only major difference for me was the build script which I don't use much anyway. 

After this test I also removed the jquery imports from both customizer.js and main.js, since webpack autoloads that to every file anyway. That also seemed to speed up the compiling a little bit as well. I hope this helps!

apintocr:
Actually this is amazing.
I'd say many of Sage 9 devs are also working with Laravel, this will put the workflows even closer together.

I've tested and it worked without any issues out of the box.

wilburpowery:
Any ETA on this change being merged?

retlehs:
webpack 4 support has landed in #2122. laravel mix might be waiting until webpack 5? this PR is stale and might be revisited in the future (or we might consider another package similar to mix)

PR #2015, opened 2018-01-23
Updated bootstrap to release version 4.0.0

slackday:
After years of development a release version of Bootstrap 4 has landed 👍 

• Updated bootstrap to release version 4.0.0
• Updated popper.js to version 1.12.9

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -661,9 +661,9 @@
       }
     },
     "bootstrap": {
-      "version": "4.0.0-beta.2",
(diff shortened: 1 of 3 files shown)


retlehs:
thanks for the PR! also need it updated here before we merge: 

https://github.com/roots/sage-installer/blob/b6d941d49dd6730b0e1bc428638611d51ebce7d0/src/Presets/Bootstrap.php

slackday:
Oh I've never used the installer so wasn't aware of this place. I can arrange a PR for it if it helps you out.

Tested bootstrap 4 and popper.js on two production sites and development environment today without any issues 👍

PR #2019, opened 2018-01-30
Make 8.5.3 use the official Bootstrap 4 release

rosswintle:
I got caught out by this today. Bootstrap 4 has dropped support for Bower (which is kinda deprecated).

So when the official, non-beta BS4 docs conflicted with the BS4-beta included in Sage 8.5.3, I updated my bower.json to use the official BS release.

This then broke the build because BS4 doesn't support Bower.

So hoping you'll pull this so others using Sage 8.5 don't get the same grief I did.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### 8.5.3: August 18th, 2017
+* Popper.js now properly supports Bower ([#1953](https://github.com/roots/sage/pull/1953))
+
 ### 8.5.2: August 13th, 2017
 * Update to Bootstrap 4 Beta ([#1944](https://github.com/roots/sage/pull/1944))
 * Fix caption display ([#1878](https://github.com/roots/sage/pull/1878))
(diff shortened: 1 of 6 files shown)


retlehs:
thanks for the PR. still need bootstrap javascript to load though

see https://discourse.roots.io/t/sage-8-legacy-sites-and-bootstrap-4-stable-version/11382/9

rosswintle:
Ah. Good spot. I knew I'd miss something and I won't have found that out for a while.

Shall I update the PR? Is this something you want to add in to keep 8.5 happy?

retlehs:
yeah if you don't mind getting JS accounted for we can merge this into sage 8 and tag 8.5.4

rangiTM:
I forked Sage to work with BS4 release edition. JS etc works: https://github.com/rangiTM/sage/tree/8.5.3

rosswintle:
That's great. Thanks.  I left a comment on the commit on your fork about bower.json now being empty.

Perhaps you could create a PR for this and see if they'll merge it in? Seems like a useful next step for 8.5.

retlehs:
closing in favor of #2027 for now (has js, just needs popper.js re-added)

PR #2022, opened 2018-02-07
Remove jQuery Dependencies

asuh:
Removing jQuery dependency from main files since the rest of this is ES-2015 compatible

Document.readyState snippet added from this example:
https://www.sitepoint.com/jquery-document-ready-plain-javascript/#plainjavascriptreadyalternative

### DIFF
--- a/resources/assets/scripts/customizer.js
+++ b/resources/assets/scripts/customizer.js
@@ -1,5 +1,3 @@
-import $ from 'jquery';
-
 wp.customize('blogname', (value) => {
-  value.bind(to => $('.brand').text(to));
+  value.bind(to => querySelector('.brand').text(to));
 });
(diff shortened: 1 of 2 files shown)


QWp6t:
We decided against this and chose to use jQuery because jQuery is already a WordPress dependency and numerous WordPress plugins rely on it. That's not to mention that Sage comes with Bootstrap by default, which depends on jQuery.

See also: https://github.com/roots/sage/pull/1741

asuh:
Right, I understand. The thought was that Sage provides the option to go with no CSS framework and the `resources/assets/scripts` files are for asset building. These are the only two files in `/scripts` that require jQuery so I submitted this to make it consistent with all other files in this directory.

PR #2024, opened 2018-02-07
9.0.0

retlehs:


LeoColomb:
@retlehs So cool! Thanks!
Would be nice to bump `sage-lib` and `sage-installer` too, to be sure all these dependencies are on the same stability level (composer is sometimes too restrictive on this).

💌

samrap:
Thanks for all the amazing work! We ❤️ Sage

PR #2027, opened 2018-02-08
bootstrap 4 scss & bundled js

ecruhling:


retlehs:
thanks @ecruhling! do you mind re-adding popper.js? then we can merge this into sage 8 and tag 8.5.4

ecruhling:
bootstrap.bundle.js includes the bootstrap JS along with popper.js;

so this works fine unless you want to remove certain bootstrap JS

retlehs:
aha, i'm too used to how we've done it previously. i'm sure someone will ask about being selective with the components... we can address it then. this seems like the easiest way to go. thanks again!

PR #2034, opened 2018-02-25
Add config for Blade Directives

Log1x:
Simple PR to move Directives to a config file that is registered with `collect()`.

Config file offers an opinionated, clean, simple solution to adding directives.

- [x] Add config file for directives
- [x] Move @asset() directive to new config file
- [x] Set `trim_trailing_whitespace` to `false` for `*.md` files :eyes:

### DIFF
--- a/app/setup.php
+++ b/app/setup.php
@@ -122,7 +122,7 @@ add_action('after_setup_theme', function () {
     /**
      * Register Directives with Blade
      */
-    collect(config('directives'))->each(function ($directive, $function) {
-        sage('blade')->compiler()->directive($function, $directive);
+    collect(config('directives'))->each(function ($handler, $name) {
+        sage('blade')->compiler()->directive($name, $handler);
     });
 });


app/setup.php
```diff
@@ -120,9 +120,9 @@
     });
 
     /**
-     * Create @asset() Blade directive
+     * Register Directives with Blade
      */
-    sage('blade')->compiler()->directive('asset', function ($asset) {
-        return "<?= " . __NAMESPACE__ . "\\asset_path({$asset}); ?>";
+    collect(config('directives'))->each(function ($directive, $function) {
+        sage('blade')->compiler()->directive($function, $directive);
```
LeoColomb: OK, this is working (👍), but I think the variable are mismatched.
I mean, currently we have:
```
$directive: Closure (the function)
$function: String (the directive name)
```
and it seems more intuitive to have
```
$directive: String (the directive name)
$function: Closure (the function)
```

Right? 🙂 

(using [Laravel names](https://github.com/laravel/framework/blob/19080684fb06fd09cd0882e3b19f7847c660cec1/src/Illuminate/View/Compilers/BladeCompiler.php#L467-L477) could also be relevant here)

Log1x: I can dig it. I switched to the Laravel names as well as swapped the order. The order was a mistake when I changed from `array_map()` to `->each()` which outputs the vars oppositely.

MWDelaney:
This has been sitting a while. Do we have a good reason not to do this? Or some certain of it that doesn’t fail the tests? 👀

Log1x:
I talked with Craig about this. He wants to avoid doing anonymous functions inside of config files like this. I'm using it personally, but we can shelf this PR until acorn release.

PR #2037, opened 2018-03-07
PHP 7.1+ is now required

retlehs:
[something down the dependency tree is now requiring 7.1](https://travis-ci.org/roots/sage/jobs/350031800)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 dist: trusty
 php:
   - 7.1
-  - 7.0
   - nightly
 
 env:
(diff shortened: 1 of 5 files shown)


DavidCWebs:
Should that not be ^ 7.1.3? From problem 2 in the Travis report, but maybe I'm reading it wrong...

christianmagill:
@retlehs, Do you know if there are is any code that relies on the updated dependency? Would requiring "illuminate/support": "5.5.34" resolve the issue. I'm waiting on a server to be updated and would prefer to use 9.0.1 for the time being. Thanks!

Issue #2012, opened 2018-01-03
Unsupported use of @php directive


collegeman:
Dear amazing people at Roots;

I want to draw your attention to [this issue](https://github.com/laravel/framework/issues/22621) I reported in the Laravel project. It seems using the `@php` directive in what I've been calling the "single-tag-form" is no longer supported. The reason being that if there are any `@endphp` closing tags used in the same template, all of the content between the first `@php` and the last `@endphp` will be evaluated by the compiler to be of the same directive, resulting in unintended behavior, e.g., the intervening template code being printed to the screen.

Here's an example of the way you're using it:

https://github.com/roots/sage/blob/42db9f9c991202fda1e4939b556b85b979dab6cb/resources/views/layouts/app.blade.php#L2

This way that you're using the `@php` directive is functional, but only in isolation to the block-form that includes both the `@php` and `@endphp` tags.

roots-ladybug:
Hi @collegeman,

It looks like the issue template is missing from this issue. Please take a look at the [Contribution Guidelines](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md), which will tell you **exactly** what your ticket has to contain in order to be processable.

Please **do not** use the issue tracker for personal support requests (use the [Roots Discourse](https://discourse.roots.io/) to ask the Roots Community for help, or if you want the Roots Team to dedicate some time to your issue, we [offer our services](https://roots.io/services/) as well).

collegeman:
@roots-ladybug spanked me! 🐞

elementalhub:
Also if you're using vscode and the Laravel Blade Snippets extension, emmet won't recognize anything after a @php() directive as "blade". Therefore snippets won't work and syntax highlighting is funny, for example comment syntax will be like 
`// foo`
instead of
 `{{-- foo --}}`

(Just in case anyone stumbles upon this as well and spent as much time as I figuring out why emmet wasn't working in my blade views.)

PR #2042, opened 2018-03-17
Changed inline @php directive to full form

hambos22:
According to https://github.com/roots/sage/issues/2012 issue, Laravel 5.5 dropped support for inline `@php` directive like `@php()`. On this PR I changed every inline to full form. I've already reproduce the issue on my sage-woo fork.

### DIFF
--- a/resources/views/index.blade.php
+++ b/resources/views/index.blade.php
@@ -10,7 +10,7 @@
     {!! get_search_form(false) !!}
   @endif
 
-  @while (have_posts()) @php(the_post())
+  @while (have_posts()) @php the_post() @endphp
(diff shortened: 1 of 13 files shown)


QWp6t:
We could also recreate the previous behavior with a new directive.

```php
    /**
     * Create @inline() Blade directive
     */
    sage('blade')->compiler()->directive('inline', function ($php) {
        return "<?php $php; ?>";
    });
```

```blade
<html @inline(language_attributes())>
```

hambos22:
I've tried that already but if it has the same behavior from this https://github.com/laravel/framework/pull/20065

QWp6t:
tyvm

Issue #2043, opened 2018-03-19
Default SVGO Configuration not working


HJGreen:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Adding SVGO configuration settings to `resources/assets/build/webpack.config.optimize.js` has no effect on the production assets generated by Webpack.

**What is the expected or desired behavior?**

Changing these configuration settings should alter the generated SVG assets.

---

## Bug report

**Please provide steps to reproduce, including full log output:**

1. Add [an SVG](https://roots.io/app/themes/roots/dist/images/sage-full-white.svg) file with a title tag to the `resources/assets/images` directory
2. Modify the Imagemin SVGO configuration to disable the removeTitle rule: `removeTitle: false`
3. Run `yarn build:production`

The processed SVG file does not retain the `<title>` tag:
```svg
<svg width="257" height="128" xmlns="http://www.w3.org/2000/svg">
[...]

PR #2044, opened 2018-03-19
Fix default SVG optimisation configuration

HJGreen:
Fixes issue #2043 

Adjusts the structure of the configuration object passed to SVGO to match the expected format, as shown in the [imagemin-svgo documentation](https://github.com/imagemin/imagemin-svgo/blob/07aad157c0223e817b510432274c74b1a32ef07b/readme.md) and [svgo examples](https://github.com/svg/svgo/blob/3ec6cef333ef9106eab7f880dab9a76997101724/examples/test.js)


retlehs:
thanks a lot for both the issue and the fix 💜

Issue #2041, opened 2018-03-12
[Feature] Support dynamic requires (imports) in DOM-Router.


vkryukov76:
- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

DOM-Router is very useful pattern.

But it has one small flaw - all routes always bundled into one file that loads on every page, even when most routes will never be fired.

Webpack has a great feature called [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports).  

Unfortunately current implementation of DOM-Router works only with regular modules, not the dynamic ones.

Dynamic approach to routes can help to split some heavy routes into separate chunks that loads on-demand through only one trigger - a classname of body tag.

I already  made some changes in DOM-Router class to support `require.ensure`d and `import()`ed modules (not supported in current sage's build system now).

https://github.com/vkryukov76/sage/commit/9b41139

If upstream and community will find it useful I will be glad to submit a PR.

Log1x:
#2190

PR #2046, opened 2018-03-20
Adds support for dynamic imported modules in DOM-Router

vkryukov76:
ref #2041

### DIFF
--- a/resources/assets/scripts/main.js
+++ b/resources/assets/scripts/main.js
@@ -8,7 +8,21 @@ import "./autoload/**/*"
 import Router from './util/Router';
 import common from './routes/common';
 import home from './routes/home';
-import aboutUs from './routes/about';
+
+/**
+ * Import  the `about` route only when <body> contains 'about-us' class
+ * using dynamic import technique
+ *
+ * NOTE: if your build system supports `import()` function
+ * then you could use this construction instead:
+ * const aboutUs = () => import('./routes/about');
+ */
+
+const aboutUs = () =>  new Promise((resolve) => require.ensure('./routes/about', (require) => {
+    const about = require('./routes/about');
+    resolve(about);
+  })
+);
 
 /** Populate Router instance with DOM routes */
 const routes = new Router({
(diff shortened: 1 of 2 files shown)


swalkinshaw:
This feature would be awesome to have in Sage but I think there's two more requirements for it:

* nicer API to make it simpler to use. I don't think uses should need to write all that boilerplate since 99% of the time it would be the same?
* promises polyfill automatically included for IE support. Or at least an option to enable/disable this.

Why can't `import` work here? I don't know too much about this, but just wondering why we need to use the old Webpack specific legacy `require.ensure`

QWp6t:
From the link you posted in #2041 

> Two similar techniques are supported by webpack when it comes to dynamic code splitting. **The first and more preferable approach is to use the `import()` syntax that conforms to the ECMAScript proposal for dynamic imports.** The legacy, webpack-specific approach is to use `require.ensure`. Let's try using the first of these two approaches...

vkryukov76:
There are two issues that prevent to use modern `import()` syntax:

- First one: outdated `buble-loader` dependency. It could be resolved by simple version bump in package.json
- Second (the main one): ESlint doesn't support `import()` statement by default (see https://github.com/eslint/eslint/issues/7764) . The only way to enable it that I know is [Babel Eslint parser](https://github.com/babel/babel-eslint). Roots has switched from Babel to Buble for performance reasons, as I know. And require all Babel stuff only to support some non-standard (yet) feature seemed as a wrong approach to me.

vkryukov76:
>nicer API to make it simpler to use. I don't think uses should need to write all that boilerplate since 99% of the time it would be the same?


Unfortunately (again) as I know that's how dynamic import feature works. The modern approach is less verbose than `require.ensure` because it already returns promise.


But in both cases, we need directly pass module path as a string (not a variable). If we wrap all the magic in a function like:

```js
export default (route) => new Promise((resolve) => require.ensure(route, (require) => {
    resolve(require(route));
  })
);
```

It will not work because Webpack will not understand that he should split code into chunks.

retlehs:
thank you for the PR! i tried to get this to work to see if we could sneak it into sage 9 but didn't have luck

i'm going to close this out because @austinpray and @QWp6t will be working on this as part of sage 10/#2122

PR #2059, opened 2018-04-18
Add version to main.js and main.css

diego-betto:


knowler:
This is unnecessary since the files are already being versioned on the production build. Run `yarn build:production` to see what I mean.

diego-betto:
Yes, it's correct. But when using `yarn build` in remote development server many times browser is not requesting correct version.

QWp6t:
You can permanently enable cache-busting like this:

**resources/assets/config.json**
```json
{
  "enabled": {
    "cacheBusting": true
    }
}
```

If you want to add logic to it, then edit it in [config.js](https://github.com/roots/sage/blob/master/resources/assets/build/config.js#L27)

diego-betto:
Awesome, thank you!

PR #2060, opened 2018-04-19
Remove php blade shorthand

ttamnedlog:
Looks like a few leftover shorthand php directives.


retlehs:
thank you!

pi-mont:
Just wondering, is there a reason we shouldn't be using PHP shorthand directives in our templates?
Thanks

Log1x:
It annoyingly got deprecated. :(

retlehs:
see #2012 #2042

garrettw:
See https://github.com/roots/sage/pull/2042

PR #2070, opened 2018-05-13
Add uglifyjs plugin

jasonbaciulis:
Simply use UglifyJs plugin instead of Webpack built-in version which allows us to configure it. I'm including one useful feature to drop console.log statements to avoid eslint errors on `build:production`

Related to discourse discussion: https://discourse.roots.io/t/uglifyjs-error-on-production-build/12419/4?u=jasonbaciulis

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,7 +32,7 @@ module.exports = {
     ]
   },
   "rules": {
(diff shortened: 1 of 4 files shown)


QWp6t:
Thanks!

PR #2074, opened 2018-05-27
Proposal: more compatible comments template

lucagentile:
Use of `get_comments_number()` instead of `have_comments()`, because the last one as per documentation:

> will always return "false" until after comments_template has been called. If you need to check for comments before calling comments_template, use get_comments_number instead
[Source](https://codex.wordpress.org/Function_Reference/have_comments)

I don't know why in my project calling `include('partials.comments')` always return false.

In that kind of situation, is also necessary to pass `get_comments` as second argument to `wp_list_comments`

### DIFF
--- a/resources/views/partials/comments.blade.php
+++ b/resources/views/partials/comments.blade.php
@@ -5,13 +5,13 @@ if (post_password_required()) {
 @endphp
 
 <section id="comments" class="comments">
-  @if (have_comments())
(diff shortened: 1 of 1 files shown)


retlehs:
> I don't know why in my project calling include('partials.comments') always return false.

maybe debug further? 

> In that kind of situation, is also necessary to pass get_comments as second argument to wp_list_comments

i don't think this makes sense to be the default in sage?

PR #2075, opened 2018-05-28
Proposal: Set removeViewBox to 'false' in webpack's optimization settings

E-VANCE:
Disable removal of the SVG viewBox by default to improve compatibility. See this Discourse thread: https://discourse.roots.io/t/svg-size-change-upon-yarn-build-production/12320


retlehs:
thank you!

PR #2076, opened 2018-05-29
Normalize and enforce single quotes

rbraunschweig:


retlehs:
thanks! can you please remove the addition of `.stylelintrc.js` and keep those settings in `package.json`?

LeoColomb:
`.stylelintrc.js` is actually a good practice. Keeping those settings in `package.json` make them harder to find and read, and even worst to customize locally.

By the way, in both `*rc.js` files, object identifiers don't need at all quotes, they are not JSON.

```json
{
  "json": "quotes"
}
```

```js
export {
  js: 'no quote'
}
```

🙂

rbraunschweig:
I was also under the impression that keeping our linting configs separate from `package.json` would lead to better maintainability. We also have separate PHP and JS linting configs, so it seems only natural that the style config is its own file too.

@LeoColomb, I contemplated leaving identifiers unquoted, but a few of them have hyphens or slashes, so I went with an all-or-nothing approach purely for consistency.

retlehs:
sorry for the wait & thank you for the PR. i updated it to remove the single quotes for styles

Issue #2077, opened 2018-06-05
Postcss's control comments not working on production


Roywcm:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Postcss provides a functionality to disable specific autoprefix behavior with Control Comments. 
This is working perfect when using the yarn start command, but not with the yarn build:production command.

**What is the expected or desired behavior?**

Disable autoprefixing for rules with postcss control comments on production compiling.

---

## Bug report

**Please provide steps to reproduce, including full log output:**

1. add a rule that normaly will be auto prefixed with the control comment above. e.g.
```
image {
    /* autoprefixer: off */
    clip-path: url(#imagemask);
  }
```

2. Run `yarn start`: now it's working, without prefix
3. Run `yarn build:production`; Now it's not working with prefix.

**Please describe your local environment:**

WordPress version: 4.9.6

OS: OSX

NPM/Node version: 5.6.0/v8.8.1

**Where did the bug happen? Development or remote servers?**

Development

knowler:
Could be related:

https://discourse.roots.io/t/removing-unused-css-with-purgecss-uncss/11586/13?u=knowler

knowler:
@Roywcm I just made a PR (#2078) to provide a better solution for this. Check it out to see my findings in regards to this issue.

PR #2078, opened 2018-06-09
Enable Sass comments and run prefixing before minification

knowler:
I was looking into a fix for #2077 and here’s what I discovered:

- `cssnano` was not the issue with the above mentioned issue the desired functionality works perfectly when using normal CSS instead of Sass.
- Sass removes the source comments by default.
  - So even disabling `cssnano` completely still resulted with the same issue.
- [Autoprefixer requires `/*! autoprefixer: off */` in Sass and Less](https://github.com/postcss/autoprefixer#control-comments).

### The problem:

- When processing Sass, comments are removed before the PostCSS plugins are processed.
  - PostCSS plugins (i.e. Autoprefixer, Purgecss) often use comments for processing so we need them to be there when that processing takes place.
  - We still want `cssnano`’s `discardComments` set to `removeAll: true` since we don’t want any comments in the production built code.

### This pull request:

- Enables the source comments in Sass.
  - This is temporary since they will be removed by minification.
- Moves the autoprefixing task before minification.
  - ~I do not know the side effects of doing this. I am actively testing for anything undesirable.~ Everything seems to run fine from my tests. I do know that cssnano doesn’t remove prefixes as of `4.0.0-rc.0`, so there should be no issues there.

#### Before as described in #2077

Input:

```css
image {
  /* autoprefixer: off */
  clip-path: url(#imagemask);
}

```

Output:

```css
image{-webkit-clip-path:url(#imagemask);clip-path:url(#imagemask)}
```

#### With the important comment and discardComments removed

Input:

```css
image {
  /*! autoprefixer: off */
  clip-path: url(#imagemask);
}
```

Output:

```css
image{
  /*! autoprefixer: off */clip-path:url(#imagemask)}
```

This works but we don’t want the comment in there.

#### With this PR

Input:

```css
image {
  /*! autoprefixer: off */
  clip-path: url(#imagemask);
}
```

Output:

```css
image{clip-path:url(#imagemask)}
```

Perfect!

### Related

- #2077 
- https://discourse.roots.io/t/removing-unused-css-with-purgecss-uncss/11586/13?u=knowler

### DIFF
--- a/resources/assets/build/postcss.config.js
+++ b/resources/assets/build/postcss.config.js
@@ -8,8 +8,8 @@ module.exports = ({ file, options }) => {
   return {
     parser: options.enabled.optimize ? 'postcss-safe-parser' : undefined,
     plugins: {
-      cssnano: options.enabled.optimize ? cssnanoConfig : false,
       autoprefixer: true,
+      cssnano: options.enabled.optimize ? cssnanoConfig : false,
     },
   };
 };
--- a/resources/assets/build/webpack.config.js
+++ b/resources/assets/build/webpack.config.js
@@ -89,7 +89,12 @@ let webpackConfig = {
               },
             },
             { loader: 'resolve-url', options: { sourceMap: config.enabled.sourceMaps } },
-            { loader: 'sass', options: { sourceMap: config.enabled.sourceMaps } },
+            {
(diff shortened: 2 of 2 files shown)


retlehs:
thank you!

PR #2080, opened 2018-06-20
Dispatch event when firing routes

tormjens:
It is especially neat when importing the scripts from the parent theme in a child theme. In one of my themes I do it like this:

```js
// routes.js
export default {
    home: {
      init() {
        console.log('Home route fired.')
      }
    }
}
```

  ```js
// main.js
/**
 * Parent Themes Script
 */
import '../../../../core-theme/resources/assets/scripts/main';

/**
 * Internal modules
 */
import routes from './routes'

/**
 * Hook in to the parent themes router
 */
document.addEventListener('routed', e => {
  let route = e.detail.route,
    fn = e.detail.fn;

  if (routes[route] && typeof routes[route][fn] === 'function') {
    routes[route][fn]();
  }
});
```

### DIFF
--- a/resources/assets/scripts/util/Router.js
+++ b/resources/assets/scripts/util/Router.js
@@ -25,6 +25,14 @@ class Router {
    * @param {string} [arg] Any custom argument to be passed to the event.
    */
   fire(route, event = 'init', arg) {
+    document.dispatchEvent(new CustomEvent('routed', {
+      bubbles: true,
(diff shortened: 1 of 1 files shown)


retlehs:
thanks and sorry for the wait! 

devs needing to support ie11 will need a polyfill: https://github.com/kumarharsh/custom-event-polyfill

PR #2084, opened 2018-07-18
Use better merge method for Webpack Preset config

knowler:
Using `merge.smartStrategy` with `'module.loaders' = 'replace'` to load the preset config allows presets to override the the default webpack config. This is necessary for [the coming Tailwind preset](https://github.com/roots/sage-installer/pull/22) which requires `resolve-url-loader` and source maps to be disabled.

### DIFF
--- a/resources/assets/build/webpack.config.js
+++ b/resources/assets/build/webpack.config.js
@@ -214,4 +214,6 @@ if (config.enabled.watcher) {
   webpackConfig = merge(webpackConfig, require('./webpack.config.watch'));
 }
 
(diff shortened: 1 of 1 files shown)


knowler:
After further testing, this method doesn't work. 👀 I will look into this...

knowler:
So this does work, but just not as much as we'd like it to. `webpack-merge` can't seem to remove individual loaders. We were able to silence the warnings in roots/sage-installer#22. This PR is required for that one to merge though.

PR #2090, opened 2018-08-05
Add searchform partial and function to replace default WordPress functionality

MWDelaney:
Inspired by this thread:
https://discourse.roots.io/t/custom-search-form/9079

The WordPress searchform, and the ability to customize it, is basic WP functionality that Sage unintentionally breaks. This PR predictably replaces the WP default template with a Blade partial, much like the the comments partials we already ship, letting users use [get_search_form()](https://developer.wordpress.org/reference/functions/get_search_form/) as normal.

### DIFF
--- a/app/filters.php
+++ b/app/filters.php
@@ -84,12 +84,12 @@ add_filter('get_search_form', function () {
 /**
  * Collect data for searchform.
  */
- add_filter('sage/template/app/data', function ($data) {
-  return $data + [
-    'sf_action' => esc_url(home_url('/')),
-    'sf_screen_reader_text' => _x('Search for:', 'label', 'sage'),
-    'sf_placeholder' => esc_attr_x('Search &hellip;', 'placeholder', 'sage'),
-    'sf_current_query' => get_search_query(),
-    'sf_submit_text' => esc_attr_x('Search', 'submit button', 'sage'),
-  ];
+add_filter('sage/template/app/data', function ($data) {
+    return $data + [
+        'sf_action' => esc_url(home_url('/')),
+        'sf_screen_reader_text' => _x('Search for:', 'label', 'sage'),
+        'sf_placeholder' => esc_attr_x('Search &hellip;', 'placeholder', 'sage'),
+        'sf_current_query' => get_search_query(),
+        'sf_submit_text' => esc_attr_x('Search', 'submit button', 'sage'),
+    ];
 });
--- a/app/filters.php
+++ b/app/filters.php
@@ -84,11 +84,12 @@ add_filter('get_search_form', function () {
 /**
  * Collect data for searchform.
  */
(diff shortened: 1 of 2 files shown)


resources/views/partials/searchform.blade.php
```diff
@@ -1,7 +1,7 @@
-<form role="search" method="get" class="search-form" action="{{ esc_url( home_url( '/' ) ) }}">
+<form role="search" method="get" class="search-form" action="{{ esc_url(home_url( '/') ) }}">
```
Log1x: 👀

garrettw: See commit d93964b -- this was already done

alwaysblank:
I'm not super attached to it, I just cut it down because I felt like `search_form_` was going to lead to some very long variables names.

darrenjacoby:
Could just use `search` ?

mmirus:
Would it make more sense to put the $sf_* stuff in the app controller? Out of the box, Sage seems slightly more oriented towards using that model. Though I think the docs introduce filters as the easy way, the code itself emphasizes the controller more, and I think most discussion on Discourse focuses on that too.

No strong feelings, just raising the question. Other than that, LGTM. Thanks all!

alwaysblank:
My reason for the filter approach is that it's the "native" approach—Controller is a 3rd party package. IMO if we offer a "native" way to do a thing then our most basic implementation should use that implementation, not depend on a 3rd party package.

mmirus:
That makes sense to me. TBH I thought it was odd that Sage uses controller out of the box and not the data filters (to the extent it uses either) for the same reason. Could just be because `App::title()` wouldn't fit in a data filter.

app/filters.php
```diff
@@ -68,3 +73,22 @@
...
+/**
+ * Collect data for searchform.
+ */
+add_filter('sage/template/app/data', function ($data) {
+    $data['sf_action'] = esc_url(home_url('/'));
+    $data['sf_screen_reader_text'] = _x('Search for:', 'label', 'sage');
+    $data['sf_placeholder'] = esc_attr_x('Search &hellip;', 'placeholder', 'sage');
+    $data['sf_current_query'] = get_search_query();
+    $data['sf_submit_text'] = esc_attr_x('Search', 'submit button', 'sage');
+    return $data;
```
QWp6t:
```php
return $data + [
    'sf_action' => esc_url(home_url('/')),
    // etc...
];
```

That will preserve `sf_*` key-value pairs that are already present in the `$data` array and only merge them in if they don't already exist.

MWDelaney: Should be all set

garrettw:
Need to fix indenting to pass phpcs. Line 87 has 1 space, should be 0; the rest should use 4-space indents rather than 2.

MWDelaney:
So, yay?

PR #2091, opened 2018-08-10
Move comments functions and data to filter

MWDelaney:
All this does is move the existing comments partial functions and data to a filter, the same way we just did with the searchform.

I haven't done any work to simplify this template, or even think clearly about whether this implementation makes sense; I wanted to get the conversation started. If we're doing it this way with searchform, we should do it this way with comments.


MWDelaney:
This won't work because of how `comments_template()` works. Never mind.

mejta:
@MWDelaney 
It will work if #2100 will be merged

PR #2100, opened 2018-08-28
Add data from controller to comments template

mejta:
- Right now comments template is rendered without data from the controller. This pull request fixes that.
- Filter that renders template should have PHP_INT_MAX priority.

### DIFF
--- a/app/filters.php
+++ b/app/filters.php
@@ -72,8 +72,8 @@ add_filter('comments_template', function ($comments_template) {
         $comments_template
     );
 
-    $data = collect(get_body_class())->reduce(function ($data, $class) use ($template) {
-        return apply_filters("sage/template/{$class}/data", $data, $template);
+    $data = collect(get_body_class())->reduce(function ($data, $class) use ($comments_template) {
+        return apply_filters("sage/template/{$class}/data", $data, $comments_template);
     }, []);
 
     $theme_template = locate_template(["views/{$comments_template}", $comments_template]);
--- a/app/filters.php
+++ b/app/filters.php
@@ -84,7 +84,7 @@ add_filter('comments_template', function ($comments_template) {
     }
 
     return $comments_template;
-}, 100);
+}, PHP_INT_MAX);
 
 /**
  * Render WordPress searchform using Blade
--- a/app/filters.php
+++ b/app/filters.php
@@ -84,7 +84,7 @@ add_filter('comments_template', function ($comments_template) {
     }
 
     return $comments_template;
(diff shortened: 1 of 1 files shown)


app/filters.php
```diff
@@ -72,15 +72,19 @@
...
 
     if ($theme_template) {
-        echo template($theme_template);
+        echo template($theme_template, $data);
         return get_stylesheet_directory().'/index.php';
     }
 
     return $comments_template;
-}, 100);
+}, PHP_INT_MAX);
```
retlehs: why is this needed?

mejta: After render there should not been anything else. If there is 100 some other plugin could modify the path with priority e.g. 200 and that would result in rendering comments twice. When we have render with PHP_INT_MAX, there is a small possibility that other plugin will modify the path after render.

mejta: After render there should not been anything else. If there is 100 some other plugin could modify the path with priority e.g. 200 and that would result in rendering comments twice. When we have render with PHP_INT_MAX, there is a small possibility that other plugin will modify the path after render.

mejta: @retlehs Should I keep `PHP_INT_MAX` in here?

app/filters.php
```diff
@@ -72,10 +72,14 @@
         $comments_template
     );
 
+    $data = collect(get_body_class())->reduce(function ($data, $class) use ($template) {
```
mmirus: I think `$template` is undefined in this scope. Should be `$comments_template`?

mejta: @mmirus You are right, it's fixed.

app/filters.php
retlehs: we don't want to use `PHP_INT_MAX` in the theme /cc @QWp6t

mtx-z: What is the reason you don't want to use `PHP_INT_MAX` (for my own information) ? . Thanks.

QWp6t: I'm against it because the use of it is too hamfisted and unnecessary. I don't see a reason for us to be that aggressive with the use of that filter or any others.

mejta: Ok, it's reverted. But generally, I think that after render there should be no chance for a hook with higher priority to render comments again. It's a very same situation like here (https://github.com/roots/sage/blob/master/app/filters.php#L63) where is also render in the filter and `PHP_INT_MAX` priority. I would suggest a consistent approach for both situations, but it's not a scope of this PR and should be in different PR.

PR #2104, opened 2018-08-30
Directory structure change according to Laravel 5.7

mejta:
In Laravel 5.7 was new directory structure introduced and I like this change very very much. This pull request make this change to Sage also.

For more information see https://laravel-news.com/laravel-5-7-resources-directory-changes


knowler:
(For the future, please just make one PR. We can always squash and merge commits if you need to modify the changes.)

🤔 I'm not sure where I land with this. I love Laravel, but I find the `assets` directory helpful for Sage since it has a handful of extra (WordPress required) files in `resources/` as well as the extra `build/` + `config.json` in `assets/` (we don't have the benefit/use of a `public` directory). All of that being said, I do see benefit in having the assets one level higher as `scripts/` and `styles/` (and sometimes `images/` can) have sub-directories which could be harder to find for newcomers. Maybe I'll warm up to it. ¯\\\_(ツ)_/¯

Compare:

1. Sage https://github.com/roots/sage/tree/master/resources
2. Laravel 5.7 https://github.com/laravel/laravel/tree/develop/resources
2. Proposed changes https://github.com/mejta/sage/tree/new-resources-structure/resources

As much as it is nice to mirror the Laravel-way of doing things, I don't think it is a guiding principle for the project (and I'm not saying that you are — I'm just processing this openly). Laravel has been and continues to be helpful is serving as a model, modern PHP application for Roots' PHP projects to follow, but following their lead should be and is a suggestion.

mejta:
I don't like deep directory structure, and assets folder was for me without any additional benefit. I keep moving the content of assets folder in my projects level-up because it's more practical than having one more level. When I saw the change in Laravel project itself, I thought that it would be feasible to do this change in Sage also. Personally, I don't see the benefits of keeping assets folder you mentioned. 

I understand that this change is not a minor, but I believe that developers would like it as I am.

I also understand your thoughts, and there is no one right opinion about this. Please consider my pull request as one of the possible ways how to structure the project. But of course, I would be delighted to see that merged, because it would save me some minutes when starting a new project :)

MWDelaney:
I feel like changing this again is a recipe for confusion. Our docs and our community are both now familiar with the Sage 9 structure, as it differs from Sage 8. Changing this again mid-Sage-9 cycle would be confusing.

mejta:
@MWDelaney Fair enough and I agree. That would lead to confusion.

PR #2106, opened 2018-09-01
best design

ztvMark:
an utf-8 icon is added,
a display block is added,
id #searchform for easy editing with css,
placeholder is added with translation option

### DIFF
--- a/resources/views/partials/searchform.blade.php
+++ b/resources/views/partials/searchform.blade.php
@@ -1,7 +1,12 @@
-<form role="search" method="get" class="search-form" action="{{ $sf_action }}">
-  <label>
-    <span class="screen-reader-text">{{ $sf_screen_reader_text }}</span>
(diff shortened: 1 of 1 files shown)


darrenjacoby:
I would prefer to keep it minimal, the classes are I'm assuming from Bootstrap, which not everyone uses when building themes with Sage. Is there a reason for inline CSS?

ztvMark:
yes, the classes are bootstrap and those that come by default in styles components

mmirus:
My preference would also be to keep it minimal, and particularly I'm not fond of the inline styles there. Also, not an accessibility expert, but using only the icon as the button value seems like a step backwards.

On the other hand, using `<button>` instead of `<input>` for the submit element is arguably a bit better.

alwaysblank:
This PR introduces several significant changes to the current design with no clear benefit or explanation, so I'm closing it. Some of these changes (`<input>` to `<button>`) may deserve additional discussion, but a separate, more focused PR for those changes and discussion of them would be more ideal.

PR #2107, opened 2018-09-01
Add size of logo

ztvMark:
Setting up theme logo with custom size
https://developer.wordpress.org/reference/functions/add_theme_support/#custom-logo


alwaysblank:
I agree that this seems like unnecessary feature creep. I'm not sure what purpose it serves, and (as illustrated by this PR) adding this functionality is trivial for anyone that needs it. I don't feel that the code and PR description submitted make a good argument for why this should be part of Sage.

darrenjacoby:
Also not in favour of such feature creep.

alwaysblank:
This PR represents unnecessary feature creep and doesn't appear to advance the goals of the project, so I'm closing it.

PR #2117, opened 2018-09-28
Update browser-sync

strarsis:
`browser-sync` release `2.24.7` uses recent `opn` which works correctly in WSL/Bash on Windows.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sage",
-  "version": "9.0.1",
+  "version": "9.0.5",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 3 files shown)


strarsis:
@mmirus: The doesn't open for me in WSL when `browser-sync` (and `opn` it uses) is not recent.

mmirus:
Gotcha. Can you add the two lock files to this PR?

strarsis:
@mmirus: I have now re-added the recent release with `npm` and `yarn` to update the lockfiles.

retlehs:
thanks for the PR - the lockfile diffs threw me off, so i just ended up doing this along with a few more in #2127

Issue #1987, opened 2017-10-31
Add command to clear Blade cache


retlehs:
via https://discourse.roots.io/t/issue-with-yarn-run-start-after-yarn-build-production/10779/6?u=benword

> do you have any plans to make a command for clearing all the caches? In Laravel there’s the `view:clear` command and maybe there could be something similar for Sage to streamline situations like this. 

...also [maybe move blade cache](https://github.com/roots/sage/blob/8691074b29c4c09dae40f0f0c33dc2b51d9292e2/config/view.php#L32) into theme directory by default? 

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)

LeoColomb:
> also maybe move blade cache into theme directory by default?

Seems a good option, yup! Was there a specific reason to put cache into `upload/`?

webstractions:
**Re: Cache**
Currently, if you delete/replace a Sage theme _(yes, frogs do fly)_, there isn't a mechanism to remove the `uploads/cache` directory.

A configurable path would be nice. Similar to paths for assets, views, etc. One could put the blade cache inside the theme directory.

@retlehs Would you consider a command-line interface similar to Artisan? The dependency, `Illuminate\Console`, is already there via Sage Installer. 

Besides clearing blade caches, you could do things like `create:controller` or `create:view`.

mAAdhaTTah:
If that's something Sage wants to do, it should build on WP-CLI, imo.

webstractions:
@mAAdhaTTah Building on WP-CLI would require that its plugin be installed. The Console is already here, and serves a different purpose than WP-CLI.

alwaysblank:
Haven't been able to figure out how to get at `Illuminate\Console`, but I've updated my Blade generator WP-CLI plugin with some methods for clearing the cache: https://github.com/alwaysblank/blade-generate

webstractions:
@alwaysblank Its being used in sage-installer for configuration, theme meta, and style presets. For some reason though, it does not work on Windows machines. There is a fix for that that Taylor Otwell used in Artisan.

alwaysblank:
[...]

PR #2121, opened 2018-10-20
Remove rendered Blades on theme install/uninstall

alwaysblank:
This PR would address some of the concerns in #1987. It doesn't add a command-line function for manipulating Blades, but it does make Sage a slightly kinder WordPress citizen by cleaning up Blades when installed/uninstalled, and provides some helper methods for doing that at other times. It endeavors to be relatively careful, and only delete the files it needs to delete, leaving other files alone.

The reason it doesn't examine existing Blade templates to find their rendered versions and delete _those_ is because part of the ostensible purpose is to clean up unnecessary Blades when changing themes: If the original Blade templates have been changed or removed during development, the "orphaned" rendered files would not be found.

I considered adding the a "build all Blades on install" feature, but felt that was a little too presumptuous and could theoretically lead to significantly increased load when activating a theme if there are a lot of Blades to render. I'd be happy to modify the PR to add that feature, though.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -140,3 +140,36 @@ function display_sidebar()
     isset($display) || $display = apply_filters('sage/display_sidebar', false);
     return $display;
 }
+
+/**
+ * Returns an array of SPLFileInfo objects representing all rendered Blades.
+ *
+ * @return array
+ */
+function get_rendered_blades()
+{
+    return collect(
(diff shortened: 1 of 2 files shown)


retlehs:
i know this is old, but can we re-do this for sage 10?

Log1x:
No longer needed as of #2174

PR #2122, opened 2018-10-24
Sage 10

QWp6t:
There's a new version of Sage in the pipeline. The most significant change is the introduction of Acorn. I'll release more information about Acorn as it nears completion.

TODO

- [ ] Restructure templates (@retlehs)
- [ ] Code splitting (@austinpray, @QWp6t, @swalkinshaw)
- [ ] WooCommerce compatibility (@QWp6t, @retlehs)
- [ ] API for build process (possibly extend Laravel Mix) (@QWp6t, @austinpray)
- [ ] Acorn - https://github.com/QWp6t/acorn/issues/1 (@QWp6t)
- [ ] Documentation

Sage 10 and Acorn are both still works in progress so you should definitely not start any projects with them in their current state unless you're willing to do an inordinate amount of debugging and go through the hassle of keeping up with updates as we push them.


resources/functions.php
```diff
@@ -37,15 +34,13 @@
 /**
  * Ensure dependencies are loaded
  */
-if (!class_exists('Roots\\Sage\\Container')) {
```
LeoColomb: Is this change compatible with Bedrock usage?

Log1x: yes– it is unrelated to bedrock

LeoColomb: Thanks but that's not my question. Actually it _is_ related to Bedrock, as we can see when git blaming the file (https://github.com/roots/sage/pull/1821).

The thing is, when using Bedrock, dependencies via composer are loaded at the Bedrock level, not at the Sage level. Then these lines will raise an error.

QWp6t: It won't raise an error.

Issue #2116, opened 2018-09-20
Unexpected action/filter execution order


kimhf:
## Submit a feature request or bug report

- [X] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [X] This is a bug report
- [X] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [ ] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Some notable actions, filters and events are being executed in the following order:

1. data collected from controllers
2. wp_enqueue_scripts (action)
3. the_content (filter) on main query
4. wp_head (action)
5. get_header (action)
6. get_footer (action)
7. wp_footer (action)

**What is the expected or desired behavior?**

The actions and filters being executed in the following order:

1. get_header (action)
2. wp_head (action)
3. wp_enqueue_scripts (action)
4. data collected from controllers
5. the_content (filter) on main query
6. get_footer (action)
7. wp_footer (action)

---

## Bug report

Plugin authors often expect the execution order to be the same on every theme for these key actions and filters.

One example would be Jetpack setting up a the_content filter during get_header. The result is the filter is not applied to the main content loop for a Sage theme.

We also might use template tags in controllers, so 3. party filters should be set up before collecting the data.

**Please provide steps to reproduce, including full log output:**

The issue was tracked by logging the execution order via priority 0 hooks on tracked actions.

**Please describe your local environment:**

WordPress version: 4.9.8

OS: Windows 10

NPM/Node version: 8.11.3

**Where did the bug happen? Development or remote servers?**

On all sites with incompatible plugins.

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

This issue can be hard to notice, but is related to the issues with wp_enqueue_scripts that is fixed in the template() function. My suggestion would be to replace that fix with something covering more actions/filters to increased compatibility with more plugins.

garrettw:
I can confirm that OP's expected behavior is accurate as per http://rachievee.com/the-wordpress-hooks-firing-sequence/

Issue #1857, opened 2017-03-14
Sage 9 beta - Plugin scripts not enqueued?


lengcai:
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [x] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [x] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

I can confirm the same problem encountered and discussed in the discourse below. I'm using the code from latest branch:

https://discourse.roots.io/t/sage-9-plugin-scripts-not-enqueued/8618

In my case, I am having problem with Master Slider plugin which try to enqueue_script whenever a it's shortcode is embedded in page. I can confirmed the Master Slider plugin does invoked the wp_enqueue_script, but the script is not included in the final rendered result.

The script appear at the bottom of the html when I switches to other theme.

**What is the expected or desired behavior?**

Would expect the plugin script to be included at the bottom of the html.

---

## Bug report

**Please provide steps to reproduce, including full log output:**
1. Generate sage theme using, 
    $ composer create-project roots/sage your-theme-name dev-master
2. Install and activate this theme.
3. Install Master Slider plugin
4. Create a simple slider using the Master Slider
5. Paste the shortcode into a newly created page.
6. View the page, the slider doesn't show up as it's javascripts not enqueued

  
**Please describe your local environment:**

WordPress version: 4.7.3

OS: Linux Cent OS 6.7

NPM/Node version: v6.10.0

**Where did the bug happen? Development or remote servers?**

Both Development & remote servers

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

https://discourse.roots.io/t/sage-9-plugin-scripts-not-enqueued/8618

---

## Feature Request

**Please provide use cases for changing the current behavior:**

X

**Other relevant information:**

X

astappev:
I have the same problem with "WPBakery Visual Composer" and "Visual Composer Extensions Addon" plugins.

[...]

PR #2123, opened 2018-10-27
Hook run sequence fix

kimhf:
This brings hook run sequence in line with expected WordPress theme behavior, and solves compatibility issues with plugins that relies on this expected behavior. It also moves the initial execution of get_header, wp_head and their child hooks to before the data filters. The logic in controllers will happen closer to where that logic it is expected to happen in relation to the hook system.

This solves the issue described in #2116.

### DIFF
--- a/app/filters.php
+++ b/app/filters.php
@@ -47,6 +47,15 @@ collect([
  * Render page using Blade
  */
 add_filter('template_include', function ($template) {
+    collect(['get_header', 'wp_head'])->each(function ($tag) {
+        ob_start();
+        do_action($tag);
+        $c = ob_get_clean();
+        remove_all_actions($tag);
+        add_action($tag, function () use ($c) {
+            echo $c;
+        });
+    });
     $data = collect(get_body_class())->reduce(function ($data, $class) use ($template) {
         return apply_filters("sage/template/{$class}/data", $data, $template);
     }, []);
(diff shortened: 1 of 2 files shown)


MWDelaney:
I like this. The reason for the code removed by this PR was to account for poorly written plugins that enqueue scripts mid-page. This is often the case with WooCommerce and Gravity Forms extensions . Has this PR been tested in those scenarios? See #1857 for some details of the original problem.

app/filters.php
```diff
@@ -47,6 +47,15 @@
  * Render page using Blade
  */
 add_filter('template_include', function ($template) {
+    collect(['get_header', 'wp_head'])->each(function ($tag) {
+        ob_start();
+        do_action($tag);
+        $c = ob_get_clean();
```
JulienMelissas: Can we use a variable a little bit better than "`$c`"?

kimhf:
I updated that variable name.

I have tried this with a custom WooCommerce template loader, and it works with the add-to-cart-variation.js, that was not enqueued without the old fix.

One downside with this is that custom template loaders, for example for WooCommerce, will need to be updated if the fix is moved out of template(). The question is then is it really a benefit to move those actions (get_header, wp_head) to before the data filters. I am just assuming here that the controllers code is executed when those filters run.

MWDelaney:
@QWp6t are you OK with merging this? I'd like to get it in there.

retlehs:
thanks y'all

mmirus:
Hey @astappiev - can you create an issue with any information you have on the problem you're seeing?

ale77dan:
This change caused **TranslatePress** plugin stopped working.
Check [discourse thread](https://discourse.roots.io/t/jquery-not-defined-when-using-translatepress-plugin-on-sage-theme/17599).

PR #2129, opened 2018-11-17
Dart-Sass implementation for webpack build

### LINKED ISSUE #2128: [Feature] Dart-Sass implementation for webpack build

- [x] This is a feature request
- [ ] This is a bug report

---

## Feature Request

This is a simple [dart-sass](https://github.com/sass/dart-sass) implementation for webpack request.
I've updated `sass-loader` to 7.1 version in order to use this implementation.
And I've already made the necessary changes and open a pull-request.

Please consider this upgrade in the next release.
Thanks

mscalessio:
This closes #2128


mmirus:
Thanks, @mscalessio.

I gave this a test (vanilla Sage 9 theme using the Bootstrap preset), but didn't notice any build speed increases, even after [adding the Fibers package](https://github.com/sass/dart-sass#javascript-api). How does that compare with your experience?

It could be that once Dart Sass is compiled to JS and used with Webpack, it loses its speed benefits over the other Sass versions.

Adding Fibers:

```shell
yarn add -D fibers
```

`webpack.config.js`:

```js
const sass = require('sass');
const Fiber = require('fibers');

...

{
  loader: 'sass', options: {
    implementation: sass,
    fiber: Fiber,
    sourceMap: config.enabled.sourceMaps,
  },
},
```

I also noticed that the output from Dart Sass is much leaner, mostly because source comments are omitted in favor of source maps. The size of main.css from `yarn build` dropped from 352kb to 237kb. https://www.diffchecker.com/3k3BbrNt

The results from `yarn build:produciton` are the same size, as one would hope.

Regardless, seems to work and since Dart Sass is the official version now, we probably want to switch.

resources/assets/build/webpack.config.js
```diff
@@ -92,6 +92,7 @@ let webpackConfig = {
             { loader: 'resolve-url', options: { sourceMap: config.enabled.sourceMaps } },
             {
               loader: 'sass', options: {
+                implementation: require('sass'),
                 sourceMap: config.enabled.sourceMaps,
                 sourceComments: true,
```
mmirus: I think we can drop this line since `sourceComments` aren't going to be supported in the JavaScript API for Dart Sass.

Johannestegner:
Not sure if there is any huge difference in build speed between node-sass and the dart implementation, one thing that have a huge time difference though is the installation time!
Running a CI job with `node-sass` takes a few MINUTES more than using `sass`, this is of course if you don't have the packages cached, but still, its a big difference and the `sass` package is the recommended package!

mmirus:
Thanks for the feedback, @Johannestegner.

@retlehs @QWp6t any thoughts on this change?

retlehs:
just tested this locally

besides seeing a very slightly longer build time with this implementation, i'm also seeing a ~1000ms performance hit on compiling assets during `yarn start` compared to the current `node-sass` setup 😕

adding fibers and configuring the build based on what @mmirus provided makes it about 500ms better, but still a ~500ms performance hit

i like the idea of doing this right now, but need to test more to see if anyone else is seeing the same results. we can also punt this for now because sage 10 will use laravel mix which now uses `dart-sass` (and hopefully we can figure out how to make the build process not any slower than it currently is)

Johannestegner:
Could I recommend that a small note could be added - to the current version docs - showing how to change to dart-sass? It's a very easy change, but I recon people could be a bit confused if not used to it.

Although, if there will be a switch in next release either way, this does not seem to important in my honest opinion.

retlehs:
closing since sage 10 will be using laravel mix which i believe uses dart-sass

trainoasis:
Quite later, but relevant still : Would love to be able to remove our old Sage 9 projects to dart-sass, we have all kinds of issues with node-sass. Many errors popping up while trying to do the above, so any more info appreciated.

MikeiLL:
In another project, we recently replaced node-sass with dart-sass, which I believe also removed the requirement of having ruby set up locally.

trainoasis:
Ruby requirement for Roots/Sage ? Am I missing something? :)

MikeiLL:
> Ruby requirement for Roots/Sage ? Am I missing something? :)

I'm definitely still missing plenty after a few years of using the stack. : ) 

I was just onboarding a windows user and we were getting errors, I think in grunt-sass, about issues with Ruby sass gem which lead us to https://github.com/sass/node-sass which says you should replace it with dart-sass which we did, at which point the ruby errors went away. 

So I'm not sure if node-sass wraps, or was wrapping a ruby gem or what.

I believe homebrew is written in ruby, and I wouldn't be surprised if some dependencies of Trellis/Sage depend on it, but not sure. Trellis-cli is written in Go.

Johannestegner:
> 
> So I'm not sure if node-sass wraps, or was wrapping a ruby gem or what.

You are correct, the node-sass package uses the original sass implementation, which is written in Ruby.

PR #2132, opened 2018-11-27
Do not ignore vendors for plugins

naokazuterada:
Hi, team. Thanks for great project! I appreciate it very much!

Here is my tiny suggestion, I am very happy if you can consider it.

# Suggestion

If there is a plugin containing a vendor directory([like this](https://wordpress.org/plugins/wp-nested-pages/)), it will be ignored, so why not just ignore the vendor directory in the root?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,5 @@ dist
 node_modules
 npm-debug.log
 yarn-error.log
-vendor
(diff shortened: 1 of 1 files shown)


QWp6t:
Thanks!

PR #2138, opened 2018-12-29
Sage 10

### LINKED ISSUE #2102: Switch to webpack 4 and Babel 7

- [x] This is a feature request
- [ ] This is a bug report

I would like to upgrade build tools to webpack 4 and babel 7 which will lead to faster builds. Also I feel that it's important to keep dependencies up-to-date. What do you think? I can also start this initiative with my pull request.

QWp6t:
There's a new version of Sage in the pipeline. The most significant change is the introduction of Acorn. I'll release more information about Acorn as it nears completion.

TODO

- [x] Restructure templates (@retlehs)
- [x] Code splitting (@austinpray, @QWp6t, @swalkinshaw)
- [ ] WooCommerce compatibility (@QWp6t, @retlehs)
- [ ] API for build process (possibly extend Laravel Mix) (@QWp6t, @austinpray)
- [x] Acorn - https://github.com/QWp6t/acorn/issues/1 (@QWp6t)
- [ ] Documentation

Sage 10 and Acorn are both still works in progress so you should definitely not start any projects with them in their current state unless you're willing to do an inordinate amount of debugging and go through the hassle of keeping up with updates as we push them.

Resolves #2102


ThatGuySam:
This is cool! Laravel Mix is my #1 most wanted feature out of this project right now.

Log1x:
I did some clean up/fixes for the Sage 10 branch in #2171. Opinions welcome.

As far as Mix goes, I've always been vocal about it in the past, so I'm more than willing to get the train rolling on that. 

I guess the main question is, what is being specifically desired that is outside of the scope of Mix out of the box? All I can think of is [stylelint](https://github.com/stakahashi/laravel-mix-stylelint) and [ESLint](https://github.com/rias500/laravel-mix-eslint) both of which have packages available (or we can roll our own `laravel-mix-linting` to handle both).

Outside of the above, the only change that would be necessary after the fact would be adding the leading slash when fetching the asset in Acorn's Manifest class.

brandontamm:
tentative 10.0 beta release?

retlehs:
no ETA, but by helping support our OSS projects financially (https://www.patreon.com/rootsdev) you can help get money into the hands of the people that are putting hard work into this release

.editorconfig
```diff
@@ -14,5 +14,5 @@ trim_trailing_whitespace = false
 [*.php]
 indent_size = 4
 
-[resources/views/**.php]
```
QWp6t: You can have regular `.php` files within resources/views. Those should be 2 spaces.

And blade files outside of resources/views should ... probably not be outside of resources/views, or they should be vendored.

Log1x: I'll push a change to handle `resources/views/**.php` – as far as Blade files outside of `resources/views` – there are multiple Sage-specific Gutenberg libraries popping up that are handling the views outside of `resources/views` such as:

- `app/Blocks/Example/Example.php`
- `app/Blocks/Example/views/example.blade.php`
- `app/Blocks/Example/views/example.(css,js)` 

which despite recommending separating things like this into an mu-plugin, is a somewhat clean approach to handling individual blocks without separating everything into 3 different directories when/if they are done inside of the theme.

strarsis:
🐱 two cents:
1. Apparently `buble` is replaced with "regular" `ES` loaders, right?
2. For Gutenberg support, would `body` class switching in Gutenberg editor make sense (https://discourse.roots.io/t/gutenberg-body-class-elements-by-template/15310)?
3. Will a gettext task be added for generating PO-files for l10n?
4. How will WooCommerce theme/template support be added?
5. Will (lib)sass be replaced with new CSS features (variables) and postcss for polyfilling?

knowler:
@strarsis

1: [Mix uses Babel](https://github.com/JeffreyWay/laravel-mix/blob/40eaa995485871cee55a366e1305ae945dfbc458/src/components/JavaScript.js#L66-L81)

5: Currently there have been no plans to switch from Sass to a PostCSS workflow. Also, Mix uses `dart-sass` (which is currently the primary Sass implementation).

PR #2118, opened 2018-10-15
Better Gutenberg support

retlehs:
see https://wordpress.org/gutenberg/handbook/extensibility/theme-support/

todo:

* [ ] add `add_theme_support('editor-color-palette')` based on colors from theme (auto-save the theme CSS variables to JSON ([ref](https://itnext.io/sharing-variables-between-js-and-sass-using-webpack-sass-loader-713f51fa7fa0)))
* [ ] add theme styles to block editor

PR #2144, opened 2019-01-13
Native gutenberg templating and styles

kellymears:
The ideal version of this PR replaces Sage's `_wp-classes.scss` but the reality is a bit more complicated.  Note that this PR brings the total `build:production` up to 229kb. If you compare that with the Gutenberg default styles, I believe it is a net savings.

I tried to adhere to Bootstrap standards as much as possible but it has been a couple years since I have used this framework. I imagine that there are more bootstrap-y ways to do many things within this PR, and I think it's probably best if someone who is already opinionated about usage of Bootstrap to incorporate them. I have tried to structure my work in a way that will make that process approachable.

Note that this PR also includes an additional composer dependency: `querypath/querypath`. It is utilized in `filters.php` to wrap `.alignwide` and `.alignfull` blocks using a div with the class `.gutenberg-wrap` applied. This makes it much easier to deal with these elements.

[This analysis / study](https://codepen.io/webmandesign/project/full/AVakKM/?preview_file=examples/block-max-width.html), along with many, many hours of frustrating experimentation, have led me to believe this is the correct approach. Also note the adoption of this method by Gutenberg contributor [Chris Van Patten](https://gist.github.com/chrisvanpatten/ca93abd76e62b01cdf598ee9a9bcbd08). To me, this component of the PR serves as a reasonably lightweight fix for an issue in a very unpredictable upstream.

This PR also restructures the starter template so that Gutenberg can have the full viewport. While it is possible to have Gutenberg output restricted to a container with a width less than 100vw there are issues that crop up when dealing with `.alignwide` and `.alignfull` elements. The main solution I've seen has been to utilize something like this:

``` scss
.entry-content .alignfull,
.entry-content .alignwide {
    margin-left: calc( -100vw / 2 + 100% / 2 );
    margin-right: calc( -100vw / 2 + 100% / 2 );
    max-width: 100vw;
}
```

This works pretty well but it is not 💯. As detailed in the codepen.io study above, this approach causes issues with `wp-block-cover` that have a fixed background on various mobile devices. This is why I believe the approach in this PR is something closer to a best practice and am utilizing it in my own projects.

There are still some missing pieces of functionality, but this PR is already rather broad and I do not want to FUBAR it with complexity.


darrenjacoby:
I'm going to play devils advocate here. 

Does it make sense to write these classes using Bootstrap? If the user is to select a custom framework from the Roots installer, then this CSS will be cleared. My guess would be quite a few Sage users don't use Bootstrap these days. 

The above code could get more use being written without a framework dependency. It shouldn't be hard to re-write this without a framework. I would be happy to help looking into the blocks CSS to convert.

darrenjacoby:
`@extend` can also work in ways that not all users writing CSS expect, so I would avoid using it. `@include` would be safer, but just plain CSS is even better. `@extend text-center` is no more reusable than `text-align: center`. Same with `@extend d-none`, etc. Those classes can make sense as a class utility, but not really in CSS itself imho.

kellymears:
@darrenjacoby my original PR was framework agnostic

darrenjacoby:
I feel like that's the better route.

kellymears:
That said, I’ve seen those css skills .. please feel free to take a pass by all means !

MWDelaney:
My understanding is they Sage is supposed to work with Bootstrap by default without the installer. Alternative versions of this code would need to be written for the other frameworks and added to the installer, but this is still an appropriate addition to Sage. 

Re, `@extend`, Bootstrap doesn’t have `@include` mixins for most of these. I see the criticism of `.text-center` but I think margin and padding classes, which will use the Bootstrap spacer variable, are still a good choice here.

darrenjacoby:
Yea, it may be a broader question. That is a fair amount of CSS to maintain for each framework, especially if more frameworks added in the future. CSS frameworks often overhaul a lot on each major release too. Changing the default Sage CSS to be framework agnostic would be the most robust option imho. Just my opinion though, others may feel different.

MWDelaney:
I'm happy with whatever the solution is for this, and I'm happy to believe that I went overboard on the `@extend`s. Generally I support replacing the default WordPress block styles with Sage styles to make them more extensible/replaceable.

kellymears:
I could see this decision going either way and be happy with the outcome. 

Thankfully all of the block css lives in one directory, so the process of swapping out a bespoke set of block styles written for a different framework should be as easy as overwriting a single folder. That’s what I’ll be doing, at any rate.

Regardless, having a path forward to customize the base styles is a good thing to have now. Letting the default one load and then trying to overwrite it is a huge pain with regards to selector specificity — no matter if you’re using bourbon or UIKit. It’s also non negligible in terms of network and rendering performance.

The WordPress of tomorrow handles pretty much every element of the site through the editor. At that point it’s not just a “good thing to have”, it’s essential to the task of theming.

MWDelaney:
@mmirus I presume the BS specificity because Sage is [already](https://github.com/roots/sage/blob/master/resources/assets/styles/autoload/_bootstrap.scss#L1) BS-[specific](https://github.com/roots/sage/blob/master/resources/assets/styles/components/_forms.scss#L3) by [default](https://github.com/roots/sage/blob/master/resources/assets/styles/layouts/_header.scss#L2). Granted the default styles aren't nearly as extensive as these, but they're there.

mmirus:
That's fair, but it's also intended (I believe) to be quite decoupled from any given framework, including BS. From the user's end, currently, if you select any of the frameworks, you have a similar starting point. Even if you select no framework, you're not losing much. From our end, the current overhead of maintaining the default styles for each framework is also pretty minimal, since the just give a really bare bones starting point.

If we build this in a BS-specific way and don't reproduce it for other frameworks, we're coupling ourselves much more closely to BS than is currently the case. I don't think that's a good option.

On the other hand, if we build it in a BS-specific way and want to keep parity for other frameworks (maintaining Sage's current framework independence), then I share @darrenjacoby's concerns about making these styles BS-specific:

- they have to be reproduced and maintained for every framework we want to include in the installer
- we have to update them to keep up with BS and *n* other frameworks, not just core updates

On top of that, it likely at least marginally increases the complexity of the installer. All told, that's significant overhead.

Finally, it would put people who don't want to use any built-in framework in a worse spot relative to today, unless we also provide a vanilla SASS implementation--at which point, we might as well just use that for all of them.

For these reasons, I think implementing this in a framework-agnostic way is the best option.

On the other hand, I don't see any benefit to doing it in a BS-specific way--totally possible I'm missing something there!

MWDelaney:
I think we're combining two very real, but very separate concerns here.

Sage is supposed to be Bootstrap by default by design, which is why BS styles are committed to the repository. The installer, if run, will replace those styles with framework-appropriate styles if another framework is chosen. These are kept up to date separately.

So yes, it's a lot of new Sass to maintain, and it has to be maintained across two repositories (this one, and sage-lib), but this PR is inline with Sage's design intent (Bootstrap by default)

mmirus:
I guess the way my brain is getting to grips with this question is:

1. In philosophical or design intent terms: yes, Sage is BS-by-default, but only in a very minimal way. I don't see adding vanilla SASS/CSS to solve this problem as violating that, at least not in a way that has any negative consequences.
2. In practical terms, adhering strictly to that approach in this case does not have benefits (that I can see) but it does have significant negative consequences.

Definitely not trying to knock any of the work that was done to implement this in a BS-specific way, and hope I'm not coming across like that! I just don't think it's the best move.

If there are significant practical downsides to putting vanilla SASS in Sage or practical benefits to doing it in a BS-specific way that might outweigh the downsides, I'm more than open to going that route--just not seeing those from my perspective.

I've probably added as much noise here as is helpful at this point. :) I'll let wiser heads weigh in.

darrenjacoby:
This conversation ultimately leads to if Sage should remain BS specific. I feel like Sage has matured a lot recently, and it may be worth reviewing if it should remain BS specific. I typically remove all CSS and JS from Sage and replace with my own. 

When it comes to frontend frameworks, and CSS/JS in particular, users seem to be way more opinionated. BS feels a bit like jQuery, both were dominate a few years ago, but that has since changed with more options available. 

As far as I know, Ben included reporting for the installer, having some stats may be able to assist with the correct route to take here. 

However, I still back a framework agnostic approach for Sage, and a minimal CSS implementation overall.

retlehs:
for the sake of this PR, everything should be bootstrap by default. in the installer we can replace the missing bootstrap variables w/ some sensible defaults. the stats show that bootstrap is chosen 3x more than 'none', and the other frameworks are rarely chosen (tailwind slightly beating foundation, no one really picking bulma or tachyons).

i agree with @kellymears on adding more variables to sage to help account for things

the installer will also have a new option for opt-ing out of including gutenberg styles

if we are going to include color and font-size support for gutenberg in sage, i'd like to automate it rather than using colors and font-sizes that aren't from the theme variables

from my todo on #2118:

> add `add_theme_support('editor-color-palette')` based on colors from theme (auto-save the theme CSS variables to JSON ([ref](https://itnext.io/sharing-variables-between-js-and-sass-using-webpack-sass-loader-713f51fa7fa0)))

kellymears:
While most of the requested changes have been made I have also been using the (non bootstrap version) of these rules to help update a site I did last year. I'll be pushing that code to production in the next day or so. I've learned a lot about what works and what doesn't through that process, and also have seen a couple gaps that need to addressing.

I'll get to merging that stuff with this PR later this week 👍

kellymears:
All instances of `gutenberg` in class definitions, filenames etc. have been renamed to `wp-blocks`. `gutenberg` is no longer used in the core block editor code, as it was just a project name and is now obsolete. 

I also tried to further reduce usage of the `>` selector. I had gone a little overboard with specificity on my first pass; this causes problems with specially aligned blocks when you begin nesting them and is just generally bad form.

kellymears:
Here is a script that adds a @wordpress/hooks call to wrap the blocks in the same way as querypath. This action happens on save and thus is not a frontend dependency, and is thereby clearly preferable. 

I'll leave it to someone more invested in Sage's JS than me to implement appropriately. Note that this script depends on a number of wordpress scripts in order to execute properly; I've included a wp_enqueue_script that works:

https://gist.github.com/kellymears/00951f91791a7bcdfa0466e883806e25

kellymears:
I'm sure this is just the first step on a long road, but it feels like a big step. I have removed `querypath/querypath` and have bundled the necessary JS within sage to add the wrapper around full/wide alignment blocks without any additional PHP dependency or frontend burden.

*It is expected that existing full and wide aligned blocks will error out in the WordPress editor once installed. All full and wide align blocks made while this JS is in place will only be valid if it remains in place (if you switch themes, for instance).*

I have added a few more variables to make it a little easier for theme devs to surface inner parts of Sage's block editor SCSS for light-touch modification. But, it is the intent that this work be seen -- like the rest of Sage -- as a starting place for development and I hope that people feel free and empowered to modify everything within resources/assets/components/blocks. In the end there really isn't that much to it. 

Thanks to @MWDelaney for all of his contributions and testing.

mmirus:
Thanks for all your work on this, @kellymears! Still need my coffee, but wanted to clarify this:

> It is expected that existing full and wide aligned blocks will error out in the WordPress editor once installed. All full and wide align blocks made while this JS is in place will only be valid if it remains in place (they will error if you switch themes and don't bring the JS along, for instance).

Does that translate to the following?

1. Using this code breaks all full and wide blocks created before it was present
2. Using this code means all full and wide blocks created while it is present can't be used without it

If so, that might not be a problem for many sites that will be developed with a Sage theme initially and will never use a different one, but it would be a huge problem for many other projects where at some point devs will be implementing a new theme for existing content (either moving to Sage or away from it).

kellymears:
@mmirus Your assessment of my jumble is correct. If someone has been using the block editor's full/wide align already and wishes to switch to Sage, they will need to either:

1. redo the offending blocks, or;
2. install something like `querypath/querypath` to replicate what the hook does and remove the hook itself. 

For my existing sites I will be keeping `querypath/querypath` and not integrating this JS, for instance. For new sites I will be using the JS. Agreed that it is a serious consideration!

* * *

This morning, I made a change to the way templating is handled -- for the most part favoring the inclusion of style rules over extra markup in order to apply max-width and auto-margins to content. 

I understand that not everyone will be utilizing the block editor on their Sage sites, and I think it's easier for them to just remove a couple include statements from their SCSS than to dig through all the template markup to remove mentions of `wp-blocks`. Also, the inclusion of `wp-blocks` outside of the context of post/page content is a bit of a red herring, and I don't want to confuse authors by adding that classname to their site navigation wrapper, for instance. Lastly, I think it nicely surfaces `wp-blocks-contain()` as an include, and it is a great include to know about if you are using this code and want your non-editor-based components to have the same max-width as your blocks.

oxyc:
At least I don't see any problem with using something like:

```scss
.alignwide, .alignfull {
  width: 100vw;
  position: relative;
  left: 50%;
  transform: translateX(-50%);
}

.alignwide {
  max-width: 60rem;
}
```

https://codepen.io/anon/pen/XOQqwO

kellymears:
@oxyc the primary issue is the behavior of background images with that transform applied, as [mentioned in the article I referenced above](https://codepen.io/webmandesign/post/gutenberg-full-width-alignment-in-wordpress-themes).

From the article, and true in my experience as well:


> But in case of Cover Image block container there are also background styles applied. And if you test thoroughly, with "Fixed Background" block option enabled, you'll find out that you should stay away from this CSS code:

```
.alignfull {
    width: 100vw;
    margin-left: 50%;
    transform: translateX(-50%); /* -> Causes rendering bug! */
}
```

> It would trigger a rendering bug in web browsers!

> The bug is related to background-attachment: fixed; styles being applied on any CSS-transformed element. And as our CSS-transformed .alignfull class styles above would interact with our Cover Image fixed background, the bug would manifest itself as:
> - background image position rendered correctly, but image not fixed in Firefox,
> - or, the image being fixed, but not correctly positioned in Chrome, for example.

resources/assets/config.json
```diff
@@ -9,7 +9,7 @@
     ]
   },
   "publicPath": "/app/themes/sage",
-  "devUrl": "http://lab.tinypixel.test",
+  "devUrl": "http://example.com",
```
lgenter: Hey Kelly, not sure if the best practice is to comment here on this commit, but I thought I'd let you know this is a .com instead of a .test. Not that anyone's going to use the default value anyway, but thought you might want to know.

Thanks for putting all the work into this PR - looking forward to having this available in Sage.

garrettw: https://en.wikipedia.org/wiki/Example.com 😉

kellymears: Thanks for the catch! 

I should probably just add example.test to /etc/hosts and save myself future embarrassment. 😅

mmirus:
@kellymears maybe I'm overestimating how large of a problem that would be for folks in reality, but implementing this in Sage in a way that breaks content if Sage is switched to / from makes me very nervous. Curious what others think?

resources/assets/config.json
lgenter: @garrettw Good point. Although my laziness is saying that not having to change `.com` to `.test` every time I customize for a project sounds nice. ;)

kellymears:
I wouldn't feel comfortable if this discussion didn't take place 👍 

Four options?

- Use `querypath/querypath`. Safest option as it doesn't manipulate content. I haven't tested its performance but I do have this running in production and it doesn't seem appreciably slower now that it's implemented.

- Use the provided starter `wp.hooks`. Most performant, and arguably the most "WordPress" way to handle it. Also does something that I don't know can be considered a best practice -- the theme is modifying content. I could also see there being issues with plugin/future update compatibility; I'm not sure!

- Provide an auxillary/companion plugin which provides these features while also being more explicitly opt-in.

- Do nothing, leave it up to devs to implement their choice or their own solution. The downside of this is obvious: Sage isn't surfacing a solution to a relatively low level problem; devs could have been doing something way cooler with their time.

MWDelaney:
* `querypath` is appealing because I’ve already run across cases where I want to add some wrapping divs to default blocks. This would give everyone a good starting place, and example, to do this themselves. 

* `wp.hooks` is appealing because it’s more in keeping with how Gutenberg is supposed to work. It’s more pure. 

So on balance `hooks` seems like the right choice even though I personally want the functionality of `querypath` myself.

duffner:
I'm looking using Gutenberg within a new site I'm starting, and I'm curious if you have any idea when this will make it into core?

ouun:
>     * Use the provided starter `wp.hooks`. Most performant, and arguably the most "WordPress" way to handle it. Also does something that I don't know can be considered a best practice -- the theme is modifying content. I could also see there being issues with plugin/future update compatibility; I'm not sure!

I would like to come up with another "WordPress" way as there is another filter `'render_block'` added: https://github.com/WordPress/gutenberg/issues/11523#issuecomment-437851918

```
add_filter( 'render_block', function( $block_content, $block ) {

   if($block['attrs']['align'] === 'wide') {
       $block_content = '<div class="wp-block-wrap test-wrap-wide">' . $block_content . '</div>';
   }

    if($block['attrs']['align'] === 'full') {
        $block_content = '<div class="wp-block-wrap test-wrap-full">' . $block_content . '</div>';
    }

    return $block_content;
}, 10, 2 );
```

The above is a very rough example and I tested it only with a couple of blocks. But it seems to do what is required here: It keeps the saved HTML code untouched and only interrupts while rendering. So switching the theme should not be an issue with that.

Thank you all for the amazong work. I'm looking forward to get rid of `calc( -100vw / 2 + 100% / 2 )` in the best possible way.

Philipp

kellymears:
@ouun This is such an excellent and needed contribution! Thank you! 

I think this is definitely the way to go and i'll be adding something along these lines when I get a moment.

evidently i need to catch up on what new block filters are available...

resources/views/partials/content.blade.php
```diff
@@ -3,7 +3,7 @@
     <h2 class="entry-title"><a href="{{ get_permalink() }}">{!! get_the_title() !!}</a></h2>
     @include('partials/entry-meta')
   </header>
-  <div class="entry-summary">
+  <main class="entry-summary">
```
knowler: This (and the one in `partials.content-single`) should remain as `<div>` since the `layouts.app` already wraps the content in `<main>`.

> A document must not have more than one main element that does not have the hidden attribute specified.

https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element

kellymears: Will replace this; thank you for the working group reference. I've been operating under the outdated understanding that nested `<main>`s were fine (if within a semantic element like `<article>`), and that the one-per rule applied to root. But, this has clearly not been the case for like two years.

kellymears: Addressed here: https://github.com/roots/sage/pull/2144/commits/7bf9810b3b7414c85d803becbc23b18e8895f89b

MWDelaney:
I know this PR is getting bigger and bigger, but it seems to me that the one big missing feature is color pallets. @log1x had a good
Idea for how to pull this off with Sage, making the Sage `$theme-colors` array available in js where Gutenberg needs it.

I really don’t know how much more work it would be, but it would make this effort really complete.

resources/assets/styles/components/blocks/_colors.scss
```diff
@@ -0,0 +1,15 @@
...
+  .has-#{$name}-color {
+    color: $color;
+  }
+
+  .has-#{$name}-background-color::before {
+    background-color: $color;
+  }
+
+  @for $i from 1 through 10 {
+    .has-background-dim.has-background-dim-#{ $i * 10 }::before {
```
mmirus: This selector doesn't actually connect the dim bg styles to the colors. Needs to be something like:

```scss
@for $i from 1 through 10 {
  .has-#{$name}-background-color.has-background-dim-#{ $i * 10 }::before {
    background-color: rgba($color, $i * 0.1) !important;
  }
}
```

ouun:
@kellymears is there something to do one can assist. Really looking forward to see with what you'll come up at the end and of cause to use a better approach as currently. 

I am also wondering how this will be usable with Tailwind as it is currently very related to Bootstrap, isn't it?

kellymears:
> @kellymears is there something to do one can assist. Really looking forward to see with what you'll come up at the end and of cause to use a better approach as currently.

Hey @ouun thanks for the offer. I had a lot of work land and that pretty much totally wiped out the past two weeks for me. Good problem for me, but still, apologies that I haven't pushed to this. I'll have something as soon as I can :)

> I am also wondering how this will be usable with Tailwind as it is currently very related to Bootstrap, isn't it?

In its current state this implementation does not require Bootstrap, but it does utilize Bootstrap if present. I'm doing this with an SCSS mixin that checks for a sage SCSS variable that the themer can use to indicate if they are using Bootstrap:

https://github.com/roots/sage/blob/eb3caeb2ecd6894c34a153861d4daa07483a8f77/resources/assets/styles/components/blocks/_utilities.scss#L26

Then, in any of the .wp-block-* partials, I can describe the fallback to be used if bootstrap is not enabled, and pass those in to the mixin, along with the bootstrap classes that represent Sage default. I do this with %extend syntax so that only one or the other is compiled during build -- no extra weight. This doesn't really add that much overhead or maintenance burden to the SCSS and means that the block styles can be used or modified by anyone, regardless of framework, while also leveraging Bootstrap if it is present (it is the most popular Sage implementation by far). 

Example usage, currently implemented, would be for the `.wp-block-button` class:

https://github.com/roots/sage/blob/eb3caeb2ecd6894c34a153861d4daa07483a8f77/resources/assets/styles/components/blocks/_wp-block-button.scss#L2

As for Tailwind specifically -- I love it (I don't really use Bootstrap, actually, although I think it is fine)! And I have a repository where I'm trying to compile block styles for major frameworks if you would like to take a shot at doing something with Tailwind and contributing: https://github.com/kellymears/wp-block-styles-project. In the end, I think a proper Tailwind plugin would be the ultimate jam and I would love to do something like that.. but am obviously overcommitted at the moment.

Cheers!

resources/assets/styles/components/blocks/_colors.scss
kellymears: My latest commit includes this fix 👊🏼

kellymears:
Added support for blacklisting blocks, unregistering style variants for blocks, and registering new style variants for blocks. All of this is configured through a single json file in the assets dir. 

https://github.com/roots/sage/blob/57bddf9ca82a9f32551ffcda18f53f1c72b6e961/resources/assets/editor.json

This configuration file also generates the sass variable map used by the block styles themselves. That transform is handled with a Webpack plugin (based on an idea by @log1x). This means that all colors, and font-sizes defined in this config file are automatically available to both the editor and public posts and only need to be maintained in one place.

Looking ahead towards Sage10 I have bundled said Webpack plugin as a laravel-mix extension (https://github.com/kellymears/laravel-mix-json-to-sass-map) and it works fine in that context as well. So, not a dead-end despite the upcoming changes in build tooling.

This also adds some new stylesheet rules for the `core/cover` block.

greatislander:
FYI, I was trying this out and I ran into a weird Gutenberg bug with the Media & Text block and your use of the `render_blocks` filter. I opened a [ticket](https://github.com/WordPress/gutenberg/issues/15476), which was closed. It looks like the underlying issue will be addressed, but at the moment Media & Text blocks with the default (wide) alignment will not be given a wrapper as the default alignment is not parsed from the block markup.

kellymears:
@greatislander nice catch 👍 . This might be approachable by utilizing the `render_block_data` filter, which is called prior to  `render_data` and allows for parsing block data that is not reflected in the markup (I think that's what's happening here).

greatislander:
@kellymears Interesting! I'll check that out.

kellymears:
I figure a quick note is in order —

While this PR will probably not make it into Sage 9, and my attention has switched to Sage 10 for the most part, it seems to have helped a lot of people figure out how to get started. Thank you all for your kind messages both here and on the Roots discourse. 

I’d suggest leaving this PR open for at least a while longer for anybody who uses it as a reference (people still ask me questions about it from time to time). It has issues but it’s really not a bad place at all to start from. However, if anybody was waiting for this to become canonical I’m flagging that this is not currently being worked on and probably will not ultimately be a part of any mainstream release. 

My plan is to close this when Roots releases Sage 10.

GerardKetuma:
Does sage 10.0.0-dev (and hence Sage 10) support Gutenberg?

Log1x:
@GerardKetuma keep an eye on #2338 – @kellymears and I are brewing up something to add to it related to this :)

app/filters.php
```diff
@@ -26,6 +26,33 @@
...
+    // Only on the frontend and only for blocks that have an alignment
+    if (is_admin() || !isset($block['attrs']['align'])) {
+        return $block_content;
+    }
+
+    if ($block['attrs']['align']=='wide') {
+        return $block_content = '<div class="wp-block-wrap wp-block-wide-wrap">'. $block_content .'</div>';
+    } elseif ($block['attrs']['align']=='full') {
+        return $block_content = '<div class="wp-block-wrap wp-block-full-wrap">'. $block_content .'</div>';
+    }
```
dsturm: Could / should this be simplified to
```
if (in_array($block['attrs']['align'], ['wide', 'full'], true)) {
    return $block_content = '<div class="wp-block-wrap wp-block-' . $block['attrs']['align'] . '-wrap">'. $block_content .'</div>';
}
```
or alternatively should we use strict comparison?

app/helpers.php
```diff
@@ -136,3 +136,56 @@ function display_sidebar()
...
+
+/**
+ * Gather Colors
+ * @param  object config
+ * @return array colors
+ */
+function gather_colors($editorConfig)
+{
+    $colors = [];
+    foreach ($editorConfig->styles->colors as $name => $color) {
```
dsturm: If the 'editor.json' is missing (or corrupt) this throws an error. Should we check for this?

kellymears: Hey @dsturm these are all great suggestions ✨, but this PR is pretty much not going anywhere and has been left up purely as a resource for others. 

If you submit a PR to my branch I'd accept it but my last commit to this was in May of 2019.

I'm sure a lot of the CSS is super out-of-spec by now. The editor markup has changed a lot. Even the anticipated markup for full/wide align blocks, which you commented on, doesn't match up to how those blocks' actual markup anymore, I don't believe.

PR #2149, opened 2019-01-27
Fix incorrect asset paths for sub directory assets

### LINKED ISSUE #2148: Incorrect production asset paths when 2+ levels deep
- [X] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [X] This is a bug report
- [X] This request isn't a duplicate of an [existing issue](https://github.com/roots/sage/issues)
- [X] I've read the [docs](https://roots.io/sage/docs) and [NPM Debugging Guidelines post](https://discourse.roots.io/t/npm-debugging-guidelines-failed-npm-install-bower-install-or-gulp-build-read-this/3060) and followed them (if applicable)
- [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/c/sage) forums

---

**What is the current behavior?**

Given the following file structure:

```
# sage/resources/assets
├── scripts
│   └── tinymce
│       └── button.js
├── styles
│   └── tinymce
│       └── button.scss
└── config.json
```

We will get these results within `sage/dist/assets.json`:
`yarn build`: `scripts/tinymce/button.js`
`yarn build:production`: `tinymce/button.js`

As a result, your CSS and JS files in subdirectories will never be loaded on production environments as the key is incorrect e.g. `asset_path('scripts/tinymce/button.js')` only works without `:production` flag.

**What is the expected or desired behavior?**

We should get identical results regardless of which script is run:
`yarn build`: `scripts/tinymce/button.js`
`yarn build:production`: `scripts/tinymce/button.js`

---

## Bug report

**Please provide steps to reproduce, including full log output:**

1. Install fresh copies of Bedrock + Sage
2. Add a Javascript or SCSS file to a new directory in either `scripts` or `styles` (e.g. `sage/resources/assets/scripts/plugins/hello.js`)
3. Enqueue in `sage/app/setup.php` and add to `sage/resources/assets/config.json`
    ```
    # sage/app/setup.php
    wp_enqueue_script('sage/misc/plugins/hello.js', asset_path('scripts/misc/plugins/hello.js'), ['jquery'], null, true);
    ```
    ```
    # sage/resources/assets/config.json
    "misc/plugins/hello": [
      "./scripts/plugins/hello.js"
    ]
    ```
4. Run `yarn build:production`

**Please describe your local environment:**

WordPress version: any (tested on 4.9.8 and 5.0.3)

OS: Ubuntu Linux 18.04, Linux Mint 19

NPM/Node version: NPM 6.4.1, Node 11

**Where did the bug happen? Development or remote servers?**

Development

**Is there a related [Discourse](https://discourse.roots.io/) thread or were any utilized (please link them)?**

https://discourse.roots.io/t/cant-load-production-assets-located-in-sub-directory/14666

codepuncher:
Fixes #2148 

Only tested on Ubuntu Linux 18.04 with Node 11.

Source of the fix: https://github.com/webdeveric/webpack-assets-manifest/issues/10#issuecomment-304972466

Used a snippet from above and ensured the path correctly resolves.


QWp6t:
I haven't tested this, but ostensibly it looks good. They didn't have that customize() property when we initially implemented our setup. This looks a lot cleaner.

I'm not going to merge it just yet, though, because I'm concerned about the path stuff. Windows uses backslashes, and it's important that the manifest output with forward slashes (because URLs have to be forward slashes). The use of `path.sep` will look for backslashes on Windows, but it's possible that the plugin transforms all of them to forward slashes before writing. All of this is to say, I wanna make sure it works on Windows before we merge.

TangRufus:
looks good on mac

alwaysblank:
**Environment:**
Windows 10
Node 8.10.0

When I run this w/ the files and settings in the discourse thread, I get the following `assets.json`:
```
{
  "scripts/customizer.js": "scripts/customizer_4061b5e2.js",
  "scripts/main.js": "scripts/main_4061b5e2.js",
  "scripts/tinymce/tinymce/button.js": "scripts/tinymce/button_4061b5e2.js",
  "styles/main.css": "styles/main_4061b5e2.css",
  "styles/tinymce/tinymce/button.css": "styles/tinymce/button_4061b5e2.css"
}
```

It looks like @QWp6t was correct to be concerned. Because `path.sep` is looking for forward-slashes on Windows, it doesn't pick up the slashes in the keys in `config.json` and doesn't transform them correctly, so you end up with the weird double-keys above. Just changing `path.sep` to `/` appeared to fix it for me:

```js
key: `${path.dirname(originalValue).split('/').shift()}/${key}`,
```

AFAIK the only problem this could cause would be if someone wrote their *keys* with forward-slashes. but that seems unlikely? I suppose you could modify it a bit to try and split on both `/` and `\`.

codepuncher:
So we could use:

```js
key: `${path.dirname(originalValue).replace(/\\/g, '/').split('/').shift()}/${key}`,
```

And that should cover it?

alwaysblank:
FYI Sage currently uses webpack-assets-manifest v1.0. If we were to update to the latest version (3.x) the `customize` entry would need to be changed a bit to reflect the updated arguments passed to it:
```
customize(entry, original) {
  return {
    key: `${path.dirname(original.value).replace(/\\/g, '/').split('/').shift()}/${entry.key}`,
  };
},
````

Log1x:
#2172

PR #2154, opened 2019-02-16
Debug helper

eduardoarandah:
Debug Helper
---

Log variables to debug.log with: 

`\App\debug($mixed)`

Works with strings, arrays or any filetype supported by var_export
[https://secure.php.net/manual/en/function.var-export.php](https://secure.php.net/manual/en/function.var-export.php)

(Why var_export instead of var_dump? outputs valid php code, more readable)

Debug log is simply `debug.log` in your public path, but can be overwritten in wp-config: 

```
define('SAGE_DEBUG_LOG_PATH', '/yourpath/www/web1/public/wp-content/themes/sage/debugging.log');
```

In case you forget to remove it, works only when WP_DEBUG is true

Example
---

`\App\debug($myvar)`

```php
array (
  'academic' => 
  array (
    'academic' => 'Academic',
  ),
  'integrated' => 
  array (
    'integrated' => 'Integrated',
  ),
  'on-aircraft' => 
  array (
    'on-aircraft' => 'On-Aircraft',
  ),
  'simulator' => 
  array (
    'simulator' => 'Simulator',
  ),
)
```

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -136,3 +136,39 @@ function display_sidebar()
     isset($display) || $display = apply_filters('sage/display_sidebar', false);
     return $display;
 }
+
+/**
+ * Debug to debug.log in public dir or the file specified in wp-config SAGE_DEBUG_LOG_PATH
+ * @param  mixed $data String, array or any object supported by var_export
+ * @return bool  true if successful
+ */
+function debug($data) {
+
+  // Only works on debug
+  if (!WP_DEBUG) {
+    return false;
+  }
+
+  // debug.log file
+  $log_path = join(DIRECTORY_SEPARATOR, [ABSPATH, 'debug.log']);
+
+  // unless specified in wp-config
+  if (SAGE_DEBUG_LOG_PATH) {
+    $log_path = SAGE_DEBUG_LOG_PATH;
+  }
+
+  // create line
+  $string = "";
+  if (is_string($data)) {
+    $string = $data;
+  } else {
+    $string = var_export($data, true);
+  }
+
+  // add end of line
+  $line = $string . PHP_EOL;
+
+  // append to file
+  file_put_contents($log_path, $line, FILE_APPEND);
+  return true;
+}
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -153,7 +153,7 @@ function debug($data) {
   $log_path = join(DIRECTORY_SEPARATOR, [ABSPATH, 'debug.log']);
 
   // unless specified in wp-config
-  if (SAGE_DEBUG_LOG_PATH) {
+  if (defined(SAGE_DEBUG_LOG_PATH)) {
     $log_path = SAGE_DEBUG_LOG_PATH;
   }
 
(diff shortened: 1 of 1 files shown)


retlehs:
thanks for the PR! we're trying to hold off on changes to sage 9, and if you take a look at the equivalent file in sage 10 you'll see it's empty: https://github.com/roots/sage/blob/10.0.0-dev/app/helpers.php

also take a look at [acorn](https://github.com/roots/acorn) (new to sage 10) which includes a debugger

eduardoarandah:
Really? I can't find it 
is it var-dumper in illuminate/support? 

Dumping vars with `dd()` is super handy, but sometimes you don't want to interrupt page flow

retlehs:
https://github.com/roots/acorn/blob/aaabe6c342cd9712247e511dd84051c397231bce/src/Acorn/View/Composers/Debugger.php

eduardoarandah:
Allright!
what I don't see is dumping to a file, like monolog in laravel. 
(btw, I find var_dump a little confusing for the eyes, more fan of var_export)

what do you think? 
is that functionality covered for the quick dump-this-var-functionality ?

QWp6t:
I would recommend using [`error_log()`](http://php.net/manual/en/function.error-log.php) instead of `file_put_contents()`, and the default log file should be in `WP_CONTENT_DIR` instead of `ABSPATH` since that's what WordPress uses by default when `WP_DEBUG_LOG` is `true`.

That said, I think this should probably be in Bedrock as a dev dependency rather than Acorn or Sage.

chesio:
> what I don't see is dumping to a file, like monolog in laravel.

Then I'd suggest you should have a look at https://github.com/inpsyde/Wonolog. If you use Bedrock (or any other Composer-based stack), it's easy to set up and get running.

Anyway, I don't think there's a reason to have a custom debug/log functionality in Sage/Bedrock.

eduardoarandah:
@chesio wonolog looks great! but I think its purpose is different. 

I'm looking for something as quick & simple as `console.log` in javascript. 

_Why is this garbage in my screen? ..._ 
`debug($var)`
_Oh! the array is wrong.._
_* remove debug($var)_

Sometimes a `dd()` is enough... sometimes you don't want to to stop the flow. 
is there a better tool for the job?

eduardoarandah:
> 
> 
> I would recommend using [`error_log()`](http://php.net/manual/en/function.error-log.php) instead of `file_put_contents()`, and the default log file should be in `WP_CONTENT_DIR` instead of `ABSPATH` since that's what WordPress uses by default when `WP_DEBUG_LOG` is `true`.
> 
> That said, I think this should probably be in Bedrock as a dev dependency rather than Acorn or Sage.

yeah, but it's not for errors, just a small "helper" to help you dump things like console.log in the browser. 
log() would be awesome, but currently means "logarithm"

Log1x:
While I can understand why this would be useful, I am closing this as it would be better suited in a package instead of being shipped with Sage.

Also, with Sage 10 nearing– perhaps check out https://laravel.com/docs/5.8/logging which should mostly work out of the box after requiring `illuminate/log`, dropping in it's config, and adding the facade (alias) to `config/app.php`.

PR #2159, opened 2019-02-27
Switch from UglifyJS to Terser

alwaysblank:
UglifyJS is [old and bad](https://github.com/terser-js/terser#why-choose-terser), Terser is the new business.

The terser webpack plugin requires webpack 4.x, which is why I'm pointing this at 10 dev, not master.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,22 +14,22 @@
       }
     },
     "@babel/core": {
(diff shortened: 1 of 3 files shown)


retlehs:
closing since sage 10 will be using laravel mix

PR #2161, opened 2019-03-14
⚠️ Blade templates directly accessible

retlehs:
⚠️ [yesterday it was brought up](https://discourse.roots.io/t/blade-templates-directly-accessible-via-web-browser-on-server/15059) that you can view blade templates in plain-text when accessing the files directly from your browser

image

this is not ideal. sage isn't the only theme affected by this — any wordpress theme that uses blade or twig (all timber themes) have the same problem because the views are publicly accessible:

image

while this PR only will only benefit apache users, the install docs will also be updated (see below) with instructions for nginx users. trellis is being updated to include preventing access to blade and twig files out of the box.

ref https://github.com/roots/trellis/pull/1075
ref https://github.com/roots/docs/pull/180

if you use a managed wordpress host, you'll need to reach out to their support team to ask them if they can make a similar change to their server config. i've asked kinsta if they'd be able to make this change globally and waiting to hear back.

### DIFF
--- a/resources/views/.htaccess
+++ b/resources/views/.htaccess
@@ -1,0 +1,12 @@
+<FilesMatch ".+\.(blade\.php)$">
+    <IfModule mod_authz_core.c>
+        # Apache 2.4
+        Require all denied
+    </IfModule>
+    <IfModule !mod_authz_core.c>
+        # Apache 2.2
+        Order deny,allow
+        Deny from all
+    </IfModule>
+</FilesMatch>
+


retlehs:
rather than implementing this PR, we'll be adding this to sage:

https://github.com/roots/wp-blade-check

n-dawson:
Would it make sense to block direct access to any PHP file within the theme folder? I've taken that approach on a number of sites in the past before using Sage.

I can't think of any legitimate reason you'd need to directly access a PHP file within the theme folder.

Doing so would simultaneously solve the plain-text blade issue as well as errors for files in the app folder and direct access to dependencies in the vendor folder.

On the topic of the vendor folder, should the composer.json file be blocked by htaccess too? I feel there's some risk involved in providing a list of dependencies that can be found within vendor. Even with indexes turned off, you're informing malicious users of everything that can be found within the vendor dir. You're giving them a list of packages they can start probing for weaknesses.

retlehs:
replaced w/ #2163

PR #2163, opened 2019-04-05
Add WP Blade Check

Log1x:
[wp-blade-check](https://github.com/roots/wp-blade-check) was created to address the security concerns of having `*.blade.php` files viewable in plain-text when accessed directly. 

In an ideal world, we would simply intercept the request and return a 403 within the application, but unfortunately this isn't possible.

The next best thing was to simply check if the templates were accessible and display an admin notice pointing the user/developer to the docs to fix this server-side within nginx or Apache.

This check is only done during `admin_init`, is cached to avoid any impact on performance, and has the ability to recheck the status from inside of the admin notice. 

Please see the package [README](https://github.com/roots/wp-blade-check/blob/master/README.md) for more details.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,7 @@
     "composer/installers": "~1.0",
     "illuminate/support": "5.6.*",
     "roots/sage-lib": "~9.0.9",
+    "roots/wp-blade-check": "^1.0",
     "soberwp/controller": "~2.1.0"
   },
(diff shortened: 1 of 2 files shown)


Log1x:
This is getting canned for now. Perhaps it will be re-implemented with WP Health Check and added into Acorn alongside some other sane checks.

PR #2171, opened 2019-05-01
Sage 10 Cleanup

Log1x:
This PR turned into a little more than I originally anticipated, but I will follow some more opinionated changes with my reasoning and we can go from there.

### Enhancements

- Enable `preflight` ~by default~ when not in production.
  - Having it disabled by default is just asking for confusion, specifically with the auto-creation of the Acorn cache directory for compiled Blade templates.
- Revert `@php` and `@endphp` to `@php()`
  - Contrary to popular belief, `@php()` still works perfectly fine. The only caveat is it can not be mixed with `@endphp` in the same template. With the addition of Composers, not only should using `@php` be avoided at all costs, but `@php(...)` works perfectly fine when multi-lined. `@php()` is considerably more aesthetic, easier to read, cleaner, and doesn't break syntax highlighting in editors such as VSCode.
- Remove excessive linebreaks from `helpers.php`.
- Make config linebreaking uniform.
  - Single linebreak before and after docblocks– consistent with Laravel.
- Add missing space to `!` conditionals to stay consistent with PSR-1/2.
- Add `aside` markup to sidebar and wrap it in `@hasSection`.
- ~Add a document wrapper with an ID of `app`.~
  - ~Not only necessary for ARIA compliance (`role="document"`) – but useful to have in place for Vue, etc.~
- ~Add ARIA roles to the document, `main`, `aside`, `header`, `nav`, and `footer` wrappers.~
- Change the `entry-meta` conditional in `content-search.blade.php` to an `@includeWhen`.
- Add some sane linebreaks throughout the views and codebase to increase readability.
- Add (missing) docblocks to the example `Title.php` composer.
- ~Add [WP Blade Check](https://github.com/roots/wp-blade-check) to alert users of publicly viewable Blade files.~
- Bump version to `10.0.0` in `package.json` and `style.css`.
- Bump `roots/sage-installer` to `^1.6`.
- Change PHP requirement to `^7.1.3` to match Acorn (and Laravel).
- Remove deprecated `qwp6t/acorn` repository from `composer.json`.
- Add some sane optimizations to `composer.json` (courtesy of Laravel).
- Update the theme structure and sponsors in `README.md`.
- Set `trim_trailing_whitespace` to `false` for `*.md` files to allow linebreaks in markdown files.

### Bugfixes

- Add missing `form.search` view.
- Unescape `$title` where used.


QWp6t:
I'm not sure if I'm on board with wrapping the header and footer elements. ARIA role for main and aside are redundant since those are already their default roles, and when direct descendants of the body, the default roles for header and footer are banner and contentinfo, respectively. The document role is unnecessary since web pages are treated as documents by default. It would only be necessary if the parent element had a different role, such as a widget or an application.

Log1x:
That's fair. I'll scrap them.

QWp6t:
I see a mixture of `{!! get_something() !!}` vs `@php(something())`.

In resources/views/partials/comments.blade.php, you have:
```php
<li class="next">{!! get_next_comments_link(__('Newer comments &rarr;', 'sage')) !!}</li>
```
In resources/views/layouts/app.blade.php, you have:
```php
<html @php(language_attributes())>
```

I think we should be consistent here. I'm cool with either one, but I slightly lean toward `{!! ... !!}` because it's explicit that there is output and that the output isn't being filtered by Blade.

config/app.php
```diff
@@ -40,8 +40,8 @@
     | or doing any other checks to ensure the service is functional.
     |
     */
-    
-    'preflight' => false,
+
+    'preflight' => true,
```
QWp6t: As a compromise, I would say to have preflight set to WP_ENV !== 'production' or something so that by default it's not enabled in production.

composer.json
```diff
@@ -33,21 +33,22 @@
...
-      "url": "https://github.com/qwp6t/acorn"
-    }
-  ],
-  "minimum-stability": "dev",
   "require": {
-    "php": ">=7.1",
-    "roots/acorn": "dev-master"
+    "php": "^7.1.3",
+    "roots/acorn": "dev-master",
+    "roots/wp-blade-check": "^1.0"
```
QWp6t: I would rather move wp-blade-check functionality over to acorn or refactor it and turn it into an acorn dependency.

Log1x: Gotcha. I wasn't sure what we were planning on doing with it. When I originally suggested sage-lib for 9 / Acorn for 10, @retlehs was kind of leaning towards shipping it in Sage it's self so people could opt-out of using it.

config/app.php
Log1x: That works.

Log1x:
> I think we should be consistent here. I'm cool with either one, but I slightly lean toward `{!! ... !!}` because it's explicit that there is output and that the output isn't being filtered by Blade.

I agree on the consistency. I did it in this case out of somewhat personal appeal as we are stuck with using `@php(body_class())` below it (since `get_body_class()` returns as an array) and it looked a bit cleaner / smaller footprint. It was also what was used prior to getting rid of `@php()` (I think I'm the one who PR'd it).

Nonetheless I reverted it for now.

config/app.php
Log1x: I guess one afterthought I have on this is the default location of the `cache` being in `uploads` – perhaps this might be problematic for those who are just pushing updates to their base theme and not pushing the new cache up?

If so, would putting `cache` in `resources` maybe be more sane?

retlehs:
thanks for all this @Log1x! merging this now since it fixes the missing search view

we can address any other concerns in a new branch/PR (like the cache folder)

Log1x:
No problem. Sounds good.

PR #2173, opened 2019-05-02
Add npm-run-all

knowler:
This lets us run related scripts together pretty easily since we can use glob patterns (e.g. `lint:*` — `**` is supported too). You can either run scripts sequentially with `run-s` or in parallel with `run-p`.

See: https://github.com/mysticatea/npm-run-all


knowler:
It could also be helpful to break up the `build:production` command:

```
    "build:production": "run-s -c clean build:production:*",
    "build:production:mix": "cross-env NODE_ENV=production webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",

```

PR #2176, opened 2019-05-04
[Sage 10] Add clean:views npm script

knowler:
Since views are in a predictable location within the theme, this might be handy. A side effect of adding this is that generated views are now cleared on every production build, which might be a good thing. If not, it’s not hard to change.


QWp6t:
I do like this for now, but ideally at some point we'll have a wp-cli option to do this that reads the actual view cache path since it's configurable. At the moment people would need to update that path in config/view.php and package.json.

PR #2178, opened 2019-05-04
[Sage 10] Move built assets to storage/theme/assets

knowler:
I think it is easier to have 1 directory (storage) for generated files instead of 2 (storage and dist).

As far as semantics go, `public` — Laravel’s default (Mix’s too) for built files — doesn’t make sense for Sage since it isn’t the only public directory in Sage. `dist` is a bit better, however, I think it’s still a bit misleading since it’s not the only theme folder which is distributed. `storage` makes sense, since it a place for the theme to store files it creates.

Obviously, `storage/theme/assets` is rather long to write out, but that could be remedied by using a `mix.config.publicPath` helper within `webpack.mix.js`.


QWp6t:
Storage isn't just used for generated files, to be clear. It can be used to simply store stuff that doesn't belong elsewhere. Since Acorn Application can only have a single base path, the default base path is your theme (in our case, Sage) folder. So all the normal Laravel stuff involving storage will use sage/storage/*.

This has conflicts now since I merged that other PR, but I'd rather wait for input from others before merging anyway. My preference is to keep the dist folder, but I'm not super tied to it.

Log1x:
I'm kinda leaning towards `dist`.

`https://app.test/app/themes/sage/storage/theme/assets/styles/app.css` just feels a little intense to me.

knowler:
Let’s stick with `dist`.

PR #2179, opened 2019-05-05
[Sage 10] Minor syntax changes + publicPath helper name change

knowler:
* Combine watch rules using extended glob patterns (tested + works).
* Use template literals for the mix helpers. Makes it seem less cluttered since we lose a bunch of parentheses.
* `public` is a future reserved word in JS so I switched it to `out` (other suggestions welcome).

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -1,7 +1,7 @@
 const mix = require('laravel-mix');
 
 // Public path helper
-const public = path => `${mix.config.publicPath}/${path}`;
+const out = path => `${mix.config.publicPath}/${path}`;
 
 // Source path helper
 const src = path => `resources/assets/${path}`;
@@ -25,7 +25,7 @@ mix.browserSync({
   proxy: 'https://example.test',
   files: [
     '(app|config|resources)/**/*.php',
-    public`(styles|scripts)/**/*.(css|js)`,
+    out`(styles|scripts)/**/*.(css|js)`,
   ],
 });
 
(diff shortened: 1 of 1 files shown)


Log1x:
maybe `publicPath` instead of `out`?

also I kind of dig the parenthesis for readability purposes but thats just me

webpack.mix.js
```diff
@@ -1,7 +1,7 @@
 const mix = require('laravel-mix');
 
 // Public path helper
-const public = path => `${mix.config.publicPath}/${path}`;
+const publicPath = path => `${mix.config.publicPath}/${path}`;
 
 // Source path helper
 const src = path => `resources/assets/${path}`;
```
Log1x: Maybe `assetPath` to keep things uniform? And then we could perhaps just put them together under `// Paths`

webpack.mix.js
```diff
@@ -24,25 +24,22 @@ mix.setPublicPath('./dist');
...
-    'dist/styles/**/*.css',
-    'dist/scripts/**/*.js'
+    '(app|config|resources)/**/*.php',
+    publicPath`(styles|scripts)/**/*.(css|js)`,
   ],
 });
 
 // Styles
-mix.sass(src('styles/app.scss'), 'styles');
+mix.sass(src`styles/app.scss`, 'styles');
```
Log1x: Still not feeling the idea of dropping the parenthesis. It makes it a lot less natural / readable in my opinion. Would like to hear others thoughts!

webpack.mix.js
knowler: I was initially trying to keep the function names short so it would be easier than writing out `resources/assets`. I’m game with changing it though.

QWp6t:
lgtm :shipit:

PR #2182, opened 2019-05-07
[Sage 10] Replace jQuery’s .ready()

knowler:
Replacements of jQuery’s `$(document).ready()` function usually look something like this:
```js
function ready(callback) {
  if (document.readyState === 'complete' || document.readyState === 'interactive') {
    return callback();
  }
  document.addEventListener('DOMContentLoaded', callback);
}
```
`document.readyState === 'complete'` is [just for IE 8](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState#Browser_compatibility) — which we don’t support. `document.readyState === 'interactive'` is for everything IE9+, but it also [happens in the same sequence as `DOMContentLoaded`](https://html.spec.whatwg.org/multipage/parsing.html#the-end) which is [supported by everything that is relevant](https://caniuse.com/#search=DOMContentLoaded).

I think `document.addEventListener('DOMContentLoaded', () => {});` is all we need.

### DIFF
--- a/resources/assets/scripts/app.js
+++ b/resources/assets/scripts/app.js
@@ -10,14 +10,7 @@ import 'bootstrap';
 import Router from './util/Router';
 import common from './routes/common';
 import aboutUs from './routes/about';
-
-/**
- * Helper function for document readiness
- */
-function ready(fn) {
-  if (document.readyState !== 'loading') return fn();
-  document.addEventListener('DOMContentLoaded', fn);
-}
+import { ready } from './utils';
 
 /**
  * Populate the Router instance with DOM routes.
--- a/resources/assets/scripts/app.js
+++ b/resources/assets/scripts/app.js
@@ -25,4 +25,4 @@ const routes = new Router({
 /**
  * Load Events
  */
-jQuery(document).ready(() => routes.loadEvents());
+document.addEventListener('DOMContentLoaded', () => routes.loadEvents());
--- a/resources/assets/scripts/app.js
+++ b/resources/assets/scripts/app.js
@@ -12,6 +12,14 @@ import common from './routes/common';
 import aboutUs from './routes/about';
 
 /**
+ * Helper function for document readiness
+ */
+function ready(fn) {
+  if (document.readyState !== 'loading') return fn();
+  document.addEventListener('DOMContentLoaded', fn);
+}
+
+/**
  * Populate the Router instance with DOM routes.
  *
  * common – Fired on all pages.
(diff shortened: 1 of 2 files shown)


oxyc:
Note that `$(document).ready()` also checks if the event has already fired. If you loaded the script with the `async` attribute the browser could have fired the event already thus never triggering this listener

knowler:
@oxyc That’s good to note, but I don’t think we need to worry about that with Sage or WordPress since you cannot (easily) add the `async` attribute to scripts using `wp_enqueue_script()`. If users are tapping into the `script_loader_tag` filter to add `async` to their scripts, I would hope that they know how it effects their scripts.

Maybe we should consider [adding `defer` to our scripts and letting them live in `<head>`](https://flaviocopes.com/javascript-async-defer/#just-tell-me-the-best-way) (since `DOMContentLoaded` still fires with `defer`), but we would probably need to change where `manifest.js` is inlined (currently before `sage/vendor`). Something for the future maybe.

swalkinshaw:
There's solution to this described in https://stackoverflow.com/questions/9237044/async-loaded-scripts-with-domcontentloaded-or-load-event-handlers-not-being-call.

ie check for `document.readyState != 'loading'`

oxyc:
I think [autoptimize](https://wordpress.org/plugins/autoptimize/) has an option for adding `async` to scripts at least. The somewhat dangerous part is that it can cause a race condition that isn't always reproducible.

swalkinshaw:
I'm highly in favour of trying to remove this jQuery dependency but I also agree with @oxyc that we shouldn't break potential `async` scripts (even if they're uncommon or hard to use in WP).

So hopefully those simple methods from SO work 😄

resources/assets/scripts/app.js
```diff
@@ -11,6 +11,14 @@ import Router from './util/Router';
 import common from './routes/common';
 import aboutUs from './routes/about';
 
+/**
+ * Helper function for document readiness
+ */
+function ready(fn) {
+  if (document.readyState !== 'loading') return fn();
```
QWp6t: I think we should use `window.setTimeout(fn, 0)` so that it runs asynchronously, and I think we should extract this out into a utility function or just [pull it in from npm](https://github.com/nuclei/readyjs).

swalkinshaw: > extract this out into a utility function

Yes 👍 

> pull it in from npm.

No 👎 also that package doesn't use `setTimeout` itself?

PR #2187, opened 2019-05-15
Simplify webpack settings

kellymears:
This change extracts common configurable parameters from `webpack.mix.js` to a simply structured configuration file in Sage root. `webpack.mix.js` is moved to `resources/assets/build/webpack.mix.js` which maintains parity with Sage 9 and shouldn't be off-putting to those already used to the theme.

This change was inspired by comments from @QWp6t and is offered with the intent of lowering the bar of entry for new developers to get their builds up-and-running. It also, arguably, make it simpler for developers of all skill levels to make a quick modification to the defaults (common example: to add an svg folder developers need only add it to `entry.dirs`, as opposed to executing `mix` functions. 

It is still very easy to modify the build directly, if desired. A small amount of overhead has been added to the mixfile. 

I could not get @knowler's cool template literal function trick to work in this context, although I tried, and that is the sole reason for this reversion.


retlehs:
superseded by the work in progress @ #2504

PR #2188, opened 2019-05-15
Optimize CircleCI config

TangRufus:
Let's aim for less than 30 seconds.

See: https://circleci.com/gh/TangRufus/workflows/sage/tree/circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,46 +1,66 @@
 version: 2.1
 
-job-template: &job-template
-    working_directory: ~/circleci
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - composer-v1-{{ checksum "composer.json" }}
-            - composer-v1- 
-      - run: composer install -n && composer test
-      - save_cache:
-          key: composer-v1-{{ checksum "composer.json" }}
-          paths:
-            - vendor
-      - restore_cache:
-          keys:
-            - node-v1-{{ checksum "package.json" }}
-            - node-v1-
-      - run: yarn install && yarn run test && yarn run build:production
-      - save_cache:
-          key: node-v1-{{ checksum "package.json" }}
-          paths:
-            - node_modules
-
-jobs:
-  build-php73:
-    docker:
-      - image: circleci/php:7.3-stretch-node-browsers
-    <<: *job-template
-  build-php72:
-    docker:
-      - image: circleci/php:7.2-stretch-node-browsers
-    <<: *job-template
-  build-php71:
-    docker:
-      - image: circleci/php:7.1-stretch-node-browsers
-    <<: *job-template
+orbs:
+  sage:
+    executors:
+      node-10:
+        docker:
+          - image: 'circleci/node:10-stretch'
+      php-73:
+        docker:
+          - image: 'circleci/php:7.3-stretch'
+      php-72:
+        docker:
+          - image: 'circleci/php:7.2-stretch'
+      php-71:
+        docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,46 +1,66 @@
...
+          executor:
+            type: executor
+        executor: << parameters.executor >>
+        steps:
+          - run: php -v
+          - checkout
+          - restore_cache:
+              keys:
+                - composer-v1-{{ checksum "composer.lock" }}
+                - composer-v1-
```
swalkinshaw: this meant to be here?

TangRufus: Yes, it tells circleci to "fallback to using the latest cache if no exact match is found (See https://circleci.com/docs/2.0/caching/)". Doc: https://circleci.com/docs/2.0/language-php/#config-walkthrough

Log1x:
Nice. Didn't know about Orbs. I tried to Google something similar but didn't turn anything up.

TangRufus:
Discuss: The example on https://circleci.com/docs/2.0/language-php/#config-walkthrough is using:

```yaml
 - save_cache: 
          key: node-v1-{{ checksum "package.json" }}
          paths:
            - node_modules
```

However, I believe `yarn.lock` is a better choice. Any idea?

TangRufus:
@Log1x 
here you go: 
- https://circleci.com/docs/2.0/creating-orbs/#section=configuration
- https://github.com/ItinerisLtd/circleci-orbs-wordpress/tree/master/src

Log1x:
> Discuss: The example on https://circleci.com/docs/2.0/language-php/#config-walkthrough is using:
> 
> ```yaml
>  - save_cache: 
>           key: node-v1-{{ checksum "package.json" }}
>           paths:
>             - node_modules
> ```
> 
> However, I believe `yarn.lock` is a better choice. Any idea?

I can't imagine it matters too much since it's just using the checksum of the file to make a unique key. I had used a non-lock file so it'd force a cache bust if the lock file is forgotten in the PR- but I guess having it sit on lock might be better so the test will fail if a new package is introduced without a new lock file present (the cache won't bust and the new package won't get installed)

TangRufus:
> Cache is immutable on write: once a cache is written for a particular key like node-cache-master, it cannot be written to again.
>
> -- https://circleci.com/docs/2.0/caching/#writing-to-the-cache-in-workflows

If we use non-lock file checksum (allow forgotten lock files), this could be a better choice:

```yaml
# Run update so that we ignore `composer.lock`
- run: composer update -n --prefer-dist 
- save_cache:
  # Alway change cache key
  key: composer-v1-{{ checksum "composer.json" }}-{{ epoch }}
  paths:
  - vendor
```

Log1x:
> > Cache is immutable on write: once a cache is written for a particular key like node-cache-master, it cannot be written to again.
> > -- https://circleci.com/docs/2.0/caching/#writing-to-the-cache-in-workflows
> 
> If we use non-lock file checksum (allow forgotten lock files), this could be a better choice:
> 
> ```yaml
> # Run update so that we ignore `composer.lock`
> - run: composer update -n --prefer-dist 
> - save_cache:
>     # Alway change cache key
>     key: composer-v1-{{ checksum "composer.json" }}-{{ epoch }}
>     paths:
>       - vendor
> ```

Well in that case cache would always pop because `epoch` would be different every time. I think we can just stick to `checksum` on the lock files. I think it's sane to have tests fail if a proper lock file isn't pushed.

retlehs:
thank you @TangRufus!

Log1x:
Screenshot

PR #2189, opened 2019-05-17
 Replaces `setup.php` with a proper service provider implementation

kellymears:
`ThemeBaseServiceProvider.php` probably belongs in Acorn as a lower level utility. If this change is found to be desirable it could be moved, leaving only `ThemeServiceProvider` (which now extends it) in `app/Providers`.

I think the declarative style of this set of changes is an improvement for developer workflows and a net gain in terms of maintainability. It is otherwise identical in terms of its feature-set to the existing functional implementation.


config/theme.php
```diff
@@ -0,0 +1,236 @@
...
+    | are added to the same script $handle, they will be printed in the order
+    | they were added, i.e. the latter added code can redeclare the previous.
+    |
+    | @link https://developer.wordpress.org/reference/functions/wp_add_inline_script/#description
+    |
+    */
+
+    'inline_scripts' => [
+        [
+            'handle'   => 'scripts/manifest',
```
QWp6t: handle should be sage/vendor

kellymears: Corrected in (Commit 
8ffa414)[https://github.com/roots/sage/pull/2189/commits/8ffa414cbf1e51500545b42f3298c90991477e75]

oxyc:
To me this is moving too far into the Grunt configuration based programming style rather just coding.

kellymears:
From my perspective we are already tightly bound to the WordPress API and so this style of configuration only provides less annoyance, not less possibility. I don't really think it is any more or less  "restrictive" than before. It is an abstraction though; I appreciate what you are saying.

That said, I'm not particularly drawn to "coding", but I do love _developing_. The less time I have to spend fulfilling the rote chores of WordPress registration the more time I can spend making new things that are fun for me to write or generate value for the people I write for.  

There is nothing stopping anyone from augmenting or outright removing these configuration parameters and supplementing/replacing them using a service provider or even standalone procedural code if they would prefer to enqueue for themselves.

oxyc:
In my opinion configuration based programming like this just makes life more difficult by abstracting away the programming language. Unless it's as expressive as Ansible where you still spend hours reading docs.

Some example scenarios:

- If a script should load conditionally you either extend the Provider and add another config option for it or start defining script's in two locations.

- If you want to enqueue some third party script with an api key but also want to use the API key for the PHP API, you would probably duplicate the API key. To me the API key is the configuration–it's the one that might change and it makes sense to only have it defined once.

- To access the editor colors you'd use `config('theme.add_theme_support.editor-color-palette')` rather than say `config('theme.colors')`.

PR #2190, opened 2019-05-18
New DOM router + dynamic imports

QWp6t:
This PR removes our current DOM router and replaces it with [js-dom-router](https://github.com/roots/js-dom-router). It also adds support for dynamic imports along with an example.

Marked as WIP because we haven't published js-dom-router yet.


resources/assets/scripts/app.js
```diff
@@ -3,27 +3,28 @@
...
  *
- * common – Fired on all pages.
- * aboutUs – Fired on the About Us page, note the change from about-us to aboutUs (camelCase).
+ * See: {@link https://webpack.js.org/guides/code-splitting/#dynamic-imports | Dynamic Imports}
  */
-const routes = new Router({
-  common,
-  aboutUs,
-});
+const home = async () => import(/* webpackChunkName: "home-page" */ './routes/home');
```
knowler:
```suggestion
const home = async () => import(/* webpackChunkName: "scripts/home-page" */ './routes/home');
```

This would put the chunk in the scripts folder. Otherwise it all looks/works good.

QWp6t: `scripts/` or `scripts/routes/`? 🤔

knowler: Good point; the latter.

PR #2196, opened 2019-05-24
Add patreon to funding.yml

QWp6t:


QWp6t:
actually gonna move this to roots/.github

PR #2197, opened 2019-05-31
[Sage 10] Make Tailwind the default CSS framework + remove jQuery/Popper

knowler:
Some work needs to be done for cleaning up the templates, but this is a base setup for Tailwind + Laravel Mix + Purgecss.

Technically, in the current state of this PR, Sage 10 has no (non-dev) dependencies.


webpack.mix.js
```diff
@@ -58,5 +63,12 @@ if (!mix.inProduction()) {
 
 // Hash and version files in production.
 if (mix.inProduction()) {
-  mix.version();
+  mix.version()
+     .purgeCss({
```
mmirus: IIRC `laravel-mix-purgecss` is set up so that by default it only runs in production, so I don't think this *needs* to be inside the conditional. Probably no harm in it, though, and is arguably easier for users to understand.

Log1x: that's correct. haven't tested it yet, but I think that goes for [`.version()` as well](https://github.com/JeffreyWay/laravel-mix/blob/master/src/components/Version.js#L35-L37)

garrettw: I haven't been able to find anything that says that for `version()`. In fact [this doc page](https://laravel.com/docs/5.8/mix#versioning-and-cache-busting) seems to indicate the opposite.

Log1x: Just tested this for sanities sake– it does appear `.version()` isn't used outside of production which is in-line with the link I included to the code above. I suspect the doc page is similar to @mmirus's argument above of making it more-so easier to understand for the user.

mmirus: Didn't know that about `.version()`!

garrettw: Interesting. Good to know.

knowler:
This will require some discussion and revision.

The thought is to make Sage more of a blank slate style-wise. Tailwind is already so agressive with it’s reset/preflight/base styles that if we try to recreate the it like how it was with Bootstrap... well, we’d be recreating Bootstrap. This is totally do-able with Tailwind, since it is a DIY framework, but this just creates more work for the developer to clear out when they begin a project.

With that being said, I’ve stripped out all the bootstrap classes from the template (for the most part), deleted the style partials (left the directories though), and extracted the WP generated classes to [a new Tailwind plugin — `tailwind-wordpress`](https://github.com/knowler/tailwind-wordpress) — (purposely left unpublished — just telling NPM to install from my repo temporarily). This makes it easy for developers to remove the classes.

Also, I added Tailwind’s container component by default since I think that’s a good default.

On another now, we’re going to include the WordPress specific purgecss whitelist that @QWp6t had previously written (though we might try to PR [the official purgecss WP whitelist](https://github.com/FullHuman/purgecss-with-wordpress)).

Keeping both the WP generated classes + the purgecss whitelist in their own dependencies seem like a good idea for keep the boilerplate easier to manage.

webpack.mix.js
knowler: I just removed these since though it does make it clear it’s in production, we can also set `sourceMaps(false)` to do the same thing and reduce a bunch of boilerplate.

kellymears:
Thanks for this awesome work. It's really clean and works very well for me. 🍡

knowler:
Marking this as ready for review. Right now it’s temporarily referencing the updated WordPress Purgecss whitelist which I have an open PR for. As noted previously, it’s also using a Tailwind plugin for creating rules for the WordPress generated classes based on values defined in the Tailwind theme. Take a look and let me know what you all think. I’ll point out a few spots that I’d like eyes on in some review comments.

resources/views/components/alert.blade.php
```diff
@@ -0,0 +1,10 @@
+<div class="my-4 px-3 py-2 border rounded-sm
+  @switch($type ?? null)
+    @case('success') bg-green-200 text-green-800 border-green-500 @break
+    @case('warning') bg-yellow-200 text-yellow-800 border-yellow-500 @break
+    @case('error') bg-red-200 text-red-700 border-red-500 @break
+    @default bg-indigo-200 text-indigo-800 border-indigo-500
+  @endswitch
```
knowler: This whole bit may be particularly revolting or just a bit odd for those unfamiliar with this method of writing CSS/building components, however, I could not bring myself to remove the class names from the view and chuck them in a View Composer (would require Purgecss rules to be adjusted) or extract them a component as, IMO that defeats the purpose Tailwind aims to solve.

This is the one example component that was left in Sage and I think it does good job cover how to create aliased Blade/Tailwind components with multiple states since it uses a service provider and components (which saw little use in Sage 9).

I’d like to hear what other think though.

tailwind.config.js
```diff
@@ -0,0 +1,41 @@
...
+ | Combined with Blade, you’ll find you spend less time writing CSS and more
+ | time designing consistent, responsive templates which use dynamic, reusable
+ | components.
+ |
+ | Combined with Purgecss, you’ll find your production stylesheets are much
+ | smaller since only the classes used in the templates are kept.
+ |
+ | By default, we are extending Tailwind’s base theme, including the container
+ | core plugin, and using an external plugin for generating WordPress classes.
+ |
```
knowler: This whole spiel might be a bit long. I was trying to write something similar to what’s written in `webpack.mix.js`, but I thought it could be helpful to describe how Tailwind fits with Blade and Purgecss.

webpack.mix.js
```diff
@@ -41,22 +41,33 @@ mix.js(src`scripts/app.js`, 'scripts')
...
+  content: [
+    './resources/views/**/*.php',
+    './resources/assets/**/*.js',
+  ],
+  defaultExtractor: content => content.match(/[A-Za-z0-9-_:/]+/g) || [],
+
+  /*
+   | Add classes, IDs, or selectors that which do not exist in your views or
+   | scripts. By default, We have whitelisted classes generated by WordPress.
+   */
```
knowler: Again, maybe TMI, but I wanted to add some points for how the developer could use the rules since they might be baffling if left without mention.

Log1x:
We're opting to ship with Bootstrap but Tailwind will be an option (preset) more than likely out of alpha.

knowler:
@Log1x You might want to use the purgecss whitelist stuff from this PR. You can just add the [`purgecss-with-wordpress`](https://github.com/FullHuman/purgecss-with-wordpress) dependency. It was updated with classes @QWp6t had for the initial Sage 10 PR.

Log1x:
> @Log1x You might want to use the purgecss whitelist stuff from this [...]

Added in #2413

Issue #2207, opened 2019-06-18
add node_modules scss to webpack


HongPong:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary

Improve webpack to process SCSS files correctly when a library is in node_modules and the js code references a SCSS file in an import statement.

## Motivation

It was difficult for me to incorporate the AOS animation JS kit which references SCSS in an import statement correctly until I changed the line to scan node_modules correctly. This would help adding other libraries I think, after they have been added with yarn. (but I might be wrong)

## Additional context

I'm not too familiar with webpack so can't say if this would introduce unexpected issues

I have added a note to the discourse forum here. hope it helps:
https://discourse.roots.io/t/webpack-handler-for-node-modules-library-scss-fix/15868

Log1x:
Sage 9 will not be receiving any more updates. Sage 10 is moving to Mix which will fix any issues you may be having with importing vendor assets.

PR #2208, opened 2019-06-18
add node_modules to scss test

HongPong:
adds node_modules based scss files to webpack. issue #2207

I'm not very conversant in Webpack so not sure if this could cause other issues, however it solved the issue I was having with scss references in node_modules libraries.


Log1x:
#2207

PR #2214, opened 2019-06-26
Use WordPress’ Browserslist config

knowler:
A few tools we’re already using (Babel, Autoprefixer, Stylelint, ESlint) use Browserslist under the hood for their target browsers. I think explicitly setting it would be helpful for developers using Sage since it will encourage them to configure it themselves instead of relying on the invisible default setting ("defaults").

I’m using the `browserslist` key in `package.json` as the method since [it’s recommended](https://github.com/browserslist/browserslist#queries), however, we can always split this out to its own config file (`.browserslistrc`). I think the former is better since `package.json` is a heavily used file and it will make it very clear to the developer which browsers they are supporting. A file that lists dependencies also makes sense to provide a list of compile targets.

~~The query I’ve set also excludes IE 11 since I believe we were already doing so. You can see the list of supported browsers this includes:~~

~~https://browsersl.ist/?q=defaults%2C+not+IE+11~~

Something to note is that since we aren’t using the JS API for browserslist we do not need to include it as a (dev-)dependency for Sage.

A downside of explicitly adding this is that it may cause an increase in PRs in regards to what the defaults should be set to.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,7 @@
 {
   "private": true,
   "browserslist": [
-    "defaults",
-    "not IE 11"
+    "extends @wordpress/browserslist-config"
   ],
   "scripts": {
     "build": "cross-env NODE_ENV=development run-s mix",
@@ -21,6 +20,7 @@
   },
   "devDependencies": {
     "@babel/plugin-syntax-dynamic-import": "^7.2",
+    "@wordpress/browserslist-config": "^2.5.0",
     "babel-eslint": "^10.0.2",
     "browser-sync": "^2.26",
     "browser-sync-webpack-plugin": "2.2",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,9 @@
 {
   "private": true,
+  "browserslist": [
+    "defaults",
+    "not IE 11"
+  ],
   "scripts": {
     "build": "cross-env NODE_ENV=development run-s mix",
     "build:production": "cross-env NODE_ENV=production run-s clean mix",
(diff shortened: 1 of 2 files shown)


garrettw:
> A downside of explicitly adding this is that it may cause an increase in PRs in regards to what the defaults should be set to.

Definitely. With IE 11 still hanging on to 2% of the market, that's a browser we still target where I work, so I wouldn't be surprised if others feel the same.

knowler:
I guess this is actually already in Sage 9 using the following settings:

```
"browserslist": [
    "last 2 versions",
    "android 4",
    "opera 12"
],
```

[That includes these browsers.](https://browsersl.ist/?q=last+2+versions%2C+android+4%2C+opera+12)

It must have gotten clear out with the rest of `package.json` at some point.

retlehs:
there's also [`@wordpress/browserslist-config`](https://github.com/WordPress/gutenberg/tree/master/packages/browserslist-config) which currently supports ie11 (even though we don't)

[that includes these browsers](https://browsersl.ist/?q=%3E+1%25%2C+ie+%3E%3D+11%2C+last+1+Android+versions%2C+last+1+ChromeAndroid+versions%2C+last+2+Chrome+versions%2C+last+2+Firefox+versions%2C+last+2+Safari+versions%2C+last+2+iOS+versions%2C+last+2+Edge+versions%2C+last+2+Opera+versions)

ref https://make.wordpress.org/core/2017/02/28/the-new-editor-and-browser-support/

PR #2220, opened 2019-07-08
Acorn asset manager config

QWp6t:
Acorn now uses an Asset Manager to make it easier for themes and plugins to include multiple manifests. This is helpful for plugins to use their own manifests, and it's also helpful for child themes to add their own manifests while also loading the manifests of the parent theme.

Simple example usage can be found in this PR:

https://github.com/roots/acorn/pull/23

Upon merging this PR, we will need to update Sage 10 to use the latest version of Acorn as well, perhaps Acorn alpha release should be tagged before this is merged?

### DIFF
--- a/config/assets.php
+++ b/config/assets.php
@@ -4,40 +4,39 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Assets Directory URI
+    | Default Assets Manifest
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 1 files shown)


QWp6t:
we gotta make sure to update composer.lock as well. I didn't do that for the PR because I wasn't sure if we were going to tag acorn first. 😬

PR #2277, opened 2019-07-23
[Sage 10]  Fix argument name and documented type

LeoColomb:


Log1x:
Let's keep it called `$view` but we can change the param to string.

LeoColomb:
@Log1x If your prefer, but this is (a bit) confusing. The passed object is not a view at all...

Log1x:
Actually, I was wrong above. `$view` is in fact `\Illuminate\View\View` as seen [here](https://github.com/roots/acorn/blob/master/src/Acorn/View/Composer.php#L43-L66).

`$view` is merely returning its `__toString()` which is the view name.

LeoColomb:
Thanks for your answer, but I'm not sure to follow.

Your link refer to two method signatures that this PR does not affect ([`with()`](https://github.com/roots/sage/pull/2277/files#diff-44adc407306e9f5ce6dc3ab0c9d8c62aR27) is 100% valid).
In this composer class, `title()` is called [with a string explicitly](https://github.com/roots/sage/pull/2277/files#diff-44adc407306e9f5ce6dc3ab0c9d8c62aR29), and AFAIK `\Illuminate\View\View !== string`, right?

Another possibility would be:

```php
    public function with($data, $view)
    {
        return ['title' => $this->title($view)]; # <- passing the View
    }

    /**
     * Returns the post title.
     *
     * @param  \Illuminate\View\View $view
     * @return string
     */
    public function title($view)
    {
        if ($view->getName() !== 'partials.page-header') {
            return get_the_title();
        }
```

QWp6t:
@LeoColomb is correct, but I I'll be changing this API in Acorn pretty soon so this will need to get rewritten anyway.

PR #2278, opened 2019-07-25
Have nav menu return HTML when used within templating tags to echo

n-dawson:
`wp_nav_menu()` will output by default. In the header example, the nav menu is inside echo template tags so it's echoing a function that already outputs directly. I've passed in an argument to have `wp_nav_menu()` return the menu instead . 

Alternatively it might make sense to just wrap it in @php as is done for things like `the_content()`.


Log1x:
I'm afraid Sage 9 will not be seeing any more updates – but thank you for bringing this to my attention. I have set this to `false` in Sage 10.

PR #2298, opened 2019-08-28
Bump version of sage-installer

octoxan:
With this version bump, Sage will now correctly install the latest Tailwind version when you choose it as a preset during install. Currently all new Sage 9 projects would start you off with an old beta Tailwind release.


retlehs:
tracking here: #2466

Issue #2261, opened 2019-07-16
Add placeholder view composers


retlehs:
add more view composers out of the box and better demonstrate the automatic view discovery

* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)

PR #2320, opened 2019-10-03
Add more Composers, demonstrate view discovery

alwaysblank:
This commit attempts to expand the demonstration of Composers and their
capabilities in line with issue #2261. It includes examples of view
discovery, as well as some varied examples of passing values to a blade
from a composer. These are just some general ideas: I'm happy to take this in
a different direction and and/remove content based on feedback.


retlehs:
todo: move to docs (#2357)

PR #2321, opened 2019-10-06
Update a node-sass version

guimontme:
Change node-sass to working better, the last code crash with Node 12 or other new versions, read: https://github.com/sass/node-sass


Log1x:
Thanks for the PR, but Sage 9 will not be seeing any updates with Sage 10 around the corner.

PR #2331, opened 2019-10-27
Change to fully qualified param type

kimhf:
Tools like PHP Intelephense currently show the wrong type for the sage() function


Log1x:
Thanks for the PR, but Sage 9 will not be seeing any changes with Sage 10 around the corner.

PR #2338, opened 2019-11-14
Little bit of this, little bit of that...

Log1x:
## Change log

- ~~Remove default styles and scripts (we will implement a CLI for this in the near-future)~~ I'm re-adding slimmed down Bootstrap defaults, but we will still do a CLI for more presets post-alpha.
- Clean up default view markup
- Change `Title.php` to `Post.php` to suggest a more organized workflow (Supersedes #2320 – content of #2320 would be reorganized to documentation)
- Create an `App` composer to show an example of a global variable
- Add `align-wide` and `responsive-embeds` to `add_theme_support` by default
- Remove deprecated `soil-jquery-cdn` from `add_theme_support`
- Clean up docblocks project-wide
- Update linting to match the latest PSR-12 standards (e.g. the license docblocks)
- Add an appropriate `#app` wrapper to make the project ready for frameworks like Vue
- Add `laravel-mix-purgecss` for purgeCSS in production
- Add `laravel-mix-copy-watched` for asset watching & versioning
- Add `laravel-mix-wp-blocks` to handle compiling blocks
- Reduce the footprint of the default `webpack.mix.js`
- Append filetypes to asset names
- Add default block editor assets
- Add handling for the WordPress asset manifest
- Remove visual editor styling (Fixes #2269)
- Add Bootstrap back as the default framework
- Add sensible default styles and folders
- Remove unused `phpcs.xml`
- Move `sage-directives` and `sage-svg` to suggests
- Bump dependencies
- Add a nice boi as an author

## ???

- @kellymears will be handling a default `editor.js` thus it is blank at the moment.
- I prefer `public/` instead of `dist/`. thoughts? it looks nice.
- ~~I prefer `css/` & `js/` instead of `styles/` & `scripts/`. thoughts? shorter url, shorter in config, looks nice. thats what actually matters in the end, things just have to look nice. functionality is just an illusion.~~ i lose
- readme & changelog
- hi austin


app/Composers/App.php
```diff
@@ -4,15 +4,15 @@
...
+class App extends Composer
 {
     /**
      * List of views served by this composer.
      *
      * @var array
      */
     protected static $views = [
-        'components.alert'
+        '*',
```
QWp6t: Maybe instead of doing this, we could rename to Composers\Layouts\App.php and remove $views all together. Then it will automatically be included with layouts/app.blade.php.

app/setup.php
```diff
@@ -19,25 +28,20 @@
         wp_enqueue_script('comment-reply');
     }
 
-    $styles = ['styles/app.css'];
-
-    foreach ($styles as $stylesheet) {
-        if (asset($stylesheet)->exists()) {
-            wp_enqueue_style('sage/' . basename($stylesheet, '.css'), asset($stylesheet)->uri(), false, null);
-        }
-    }
```
QWp6t: we had this because when running browsersync, the css file isn't there. i haven't tested default sage 10 dev process in a while. so if this isn't needed anymore, then ok. but if we're still not writing CSS file, then we should probably leave this block in there.

Log1x: ah, yeah it's not an issue anymore.

app/Composers/App.php
Log1x: I considered that, but resorted to this for simplicities sake as to not complicate anything with more directories, etc. – I figured we would express the additional ways Composers can be used/auto-detected within' the docs.

alwaysblank:
I prefer `styles` and `scripts` for source files because if the directories are called `css` and `js` then I expect to find those file extensions within, where in actuality I might find Sass, TypeScript, etc. Directory names that describe the function, not the form, seem more logical to me.

Log1x:
> I prefer `styles` and `scripts`
Yeah I'm okay with either, really. I was just throwing it out there again. I understand your reasonings, and there are no arguments against them. I just don't find anything about it personally off-putting and saw it as an attractive change in Laravel. Indeed, an opinionated one.

Log1x:
> `webpack.mix.js` should probably be formatted according to the ESLint rules. Also, ASI is great and all (I used to write my JS that way), but it is considered hazardous by TC39:
> 
> > As new syntactic features are added to ECMAScript, additional cases requiring explicit semicolons emerge over time. As such, consistently explicit semicolon use is recommended.
> 
> See: https://github.com/WebReflection/about-no-semicolons

The mix config isn't linted so I wasn't thinking about it – I'll add them back shortly.

resources/assets/styles/config/variables.scss
```diff
@@ -5,7 +5,3 @@
 $theme-colors: (
   primary: #525ddc
 );
-
-/** Bootstrap Navbar Fix (https://git.io/fADqW) */
-$navbar-dark-toggler-icon-bg: none;
-$navbar-light-toggler-icon-bg: none;
```
Log1x: In case anyone is curious, this is no longer needed.

PR #2339, opened 2019-11-14
Add native support for responsive embedded content

ciromattia:
See https://developer.wordpress.org/block-editor/developers/themes/theme-support/#responsive-embedded-content


Log1x:
Thanks for the PR, but Sage 9 will not be seeing any updates as Sage 10 is right around the corner.

See [this PR](https://github.com/roots/sage/pull/2338) opened last night that includes your change.

ciromattia:
Great, didn't saw it!
Thanks :)

PR #2351, opened 2019-11-30
Sage 10 editor scripts & styles

kellymears:
Registers a block style on `core/button` and some styles for it.

To be considered against https://github.com/roots/sage/pull/2353, which is a bigger PR but also surfaces a lot of configuration methods that engender best practices.


resources/assets/styles/components/block.scss
```diff
@@ -0,0 +1,11 @@
+[class^="wp-block-"] {
```
retlehs: drop this parent? global `.has-xxxxx-color` is okay

resources/assets/styles/editor.scss
```diff
@@ -4,10 +4,25 @@
...
+    max-width: 1110px;
+  }
 
-  [example-block] {
-    @import 'components/button';
+  .editor-block-list__layout {
+    .wp-block {
+      padding-left: 43px;
+      padding-right: 43px;
+    }
```
retlehs: drop all of this?

kellymears: you holdin' stock in Noto Serif? 🤔

kellymears: 7058b40

resources/assets/styles/components/block.scss
kellymears: d2bb13d

resources/assets/styles/editor.scss
Log1x: id keep it personally. we can drop the padding, etc. but a `font-family` and `max-width` for `.wp-block` is a sane default on the editor and the font-family for `.editor-post-title__input` has to be set specifically.

just my 2 cents

kellymears: I agree that I think it super minimal and nice -- since we have a bootstrap default I feel like we're okay asserting at least some style.

Note that we can't drop the padding if we're manipulating the width like this. The explicit pixel padding looks bad but is based on the built-in editor styles (which are full of explicitly set px values). Setting it to 43px prevents the block handle from overflowing and being occluded by the sidebar.

retlehs: i think the `has-` styles within `resources/assets/styles/components/_block.scss` should be moved to `resources/assets/styles/common/global.scss`

and i think we're no longer prefixing sass files with `_` after a recent update by @Log1x 

i'd be okay with adding the font-family to the title in the editor, but let's hold off on anything else like width/padding

kellymears: https://github.com/roots/sage/pull/2351/commits/fc5a8dbd2599645a47d57fbd75559c77d1d5bef8 ✅

PR #2352, opened 2019-11-30
Add theme color palette support

kellymears:
Just adds a single theme color. 

Alternatively, we could disable theme color support.

If desired I can open a separate PR to handle disabling custom font-sizes. I think it's a good idea for them to ship disabled, personally.


kellymears:
I have submitted another PR which locks down both theme color palettes & font sizes, amongst other things https://github.com/roots/sage/pull/2353

Issue #2213, opened 2019-06-26
Upgrade node-sass to 4.12.0


mquellhorst:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

Running yarn in a fresh install results in an error for node-sass.
Upgrade to node-sass 4.12.0 fixes the issue.

**Expected behavior:**

Finish yarn without errors

**Actual behavior:**

Results in error:
`
/Users/username/.node-gyp/12.4.0/include/node/v8.h:2995:5: note: candidate constructor not viable: requires 2 arguments, but 1 was provided
    Utf8Value(Isolate* isolate, Local<v8::Value> obj);
    ^
1 error generated.
make: *** [Release/obj.target/binding/src/create_string.o] Error 1
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/Users/username/Development/sage/application/wp-content/themes/hartwerk/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (events.js:200:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:272:12)
gyp ERR! System Darwin 18.6.0
gyp ERR! command "/usr/local/Cellar/node/12.4.0/bin/node" "/Users/username/Development/sage/application/wp-content/themes/hartwerk/node_modules/node-gyp/bin/node-gyp.js" "rebuild" "--verbose" "--libsass_ext=" "--libsass_cflags=" "--libsass_ldflags=" "--libsass_library="
gyp ERR! cwd /Users/username/Development/hartwerk/application/wp-content/themes/sage/node_modules/node-sass
gyp ERR! node -v v12.4.0`

**Reproduces how often:** [What percentage of the time does it reproduce?]

Always

## Versions

sage 9.0.9
mac with osx 10.14.5
node v12.4.0

retlehs:
looks like sage 9 still works fine with the latest node LTS (version 10)

[...]

PR #2367, opened 2019-12-25
Bump node-sass to 4.12.0

ridgekuhn:
I tried installing a fresh copy of Sage yesterday and got an error after running `yarn`. [Sage docs](https://roots.io/sage/docs/existing-sage-projects/), recommend using NodeJS LTS which is currently v12. The [minimum node-sass version](https://www.npmjs.com/package/node-sass) for Node 12 is now node-sass 4.12+. Error output attached.

[node-sass.txt](https://github.com/roots/sage/files/3998983/node-sass.txt)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "imagemin-mozjpeg": "~7.0.0",
     "imagemin-webpack-plugin": "~2.2.0",
     "import-glob": "~1.5",
-    "node-sass": "~4.9.4",
+    "node-sass": "~4.12.0",
     "postcss-loader": "~2.1.0",
     "postcss-safe-parser": "~3.0",
     "resolve-url-loader": "~2.3.1",


ridgekuhn:
Just realized that since lock files are present in the repo, this won't work without regenerating them. I'm not sure if it's appropriate to regenerate those myself and push it upstream, so here's a patch with new lock files. 

[0001-Added-lock-files.patch.txt](https://github.com/roots/sage/files/4001306/0001-Added-lock-files.patch.txt)

raffjones:
This PR fixes an issue I am seeing on every attempted fresh install using node 12.8.1 and yarn 1.17.3

pfalomorblab:
This would fix this https://github.com/roots/sage/issues/2213

retlehs:
tracking here: #2466

Issue #2370, opened 2019-12-25
Using ES6 classes causes editor.js to not execute


oxyc:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

Not sure when this happened as it runs fine on an older site based on sage 10.0.0-dev. Anyways as title says, using ES6 classes causes `editor.js` to not execute. Using es6 classes in `app.js` works fine. Check _Additional information_ further down for some initial debugging.

## Steps to reproduce

1. Add the following to the `editor.js` file:

    ```js
    class Foobar {}
    console.log('test');
    ```

2. Run `npm run build`
3. Edit a page with the block editor and open the console

**Expected behavior:** See `test` logged in the console

**Actual behavior:** Nothing is logged.

**Reproduces how often:** Always

## Versions

Sage: 10.0.0-dev branch
Node v12.12.0 (same on 10.17.0)

## Additional information

I dug a bit and turns out this is caused by [`@wordpress/babel-preset-default`](https://github.com/kellymears/laravel-mix-wp-blocks/) which is added to laravel mix thorugh [`laravel-mix-wp-block`](https://github.com/kellymears/laravel-mix-wp-blocks/)

- If I remove the babel preset, the console command runs.
- If I add the class to `app.js` instead, the console command runs.

When inspecting the difference of the compiled files I noticed that `@wordpress/babel-preset-default` seem to [replace inlined babel runtime helpers](https://babeljs.io/docs/en/babel-plugin-transform-runtime#helpers) unlike the default laravel mix `.js()` command. Not sure if this is the cause though.

### Without `@wordpress/babel-preset-default` (works)

```js
"use strict";
__webpack_require__.r(__webpack_exports__);
/* harmony import */ var _wordpress_edit_post__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(/*! @wordpress/edit-post */ "@wordpress/edit-post");
/* harmony import */ var _wordpress_edit_post__WEBPACK_IMPORTED_MODULE_0___default = /*#__PURE__*/__webpack_require__.n(_wordpress_edit_post__WEBPACK_IMPORTED_MODULE_0__);
[...]

PR #2371, opened 2019-12-25
Fix vendor extracted js modules not available in editor.

oxyc:
See #2370 

I know sage has always registered the scripts at the same time they're enqueued so this solution might not be great to everyone.

Another option is to somehow fix it in the laravel mix config I guess.

### DIFF
--- a/app/setup.php
+++ b/app/setup.php
@@ -36,14 +36,15 @@ add_action('wp_enqueue_scripts', function () {
  */
 add_action('enqueue_block_editor_assets', function () {
     if ($manifest = asset('scripts/manifest.asset.php')->get()) {
+        wp_enqueue_script('sage/vendor.js', asset('scripts/vendor.js')->uri(), ['jquery'], null, true);
         wp_enqueue_script(
             'sage/editor.js',
             asset('scripts/editor.js')->uri(),
-            $manifest['dependencies'],
+            array_merge($manifest['dependencies'], ['sage/vendor.js']),
             $manifest['version']
         );
 
-        wp_add_inline_script('sage/editor.js', asset('scripts/manifest.js')->contents(), 'before');
+        wp_add_inline_script('sage/vendor.js', asset('scripts/manifest.js')->contents(), 'before');
     }
 
     wp_enqueue_style('sage/editor.css', asset('styles/editor.css')->uri(), false, null);


Log1x:
`editor.js` should not be enqueued on the frontend.

By default, I do not enqueue `vendor.js` as it is not required with the default behavior, but can simply be enqueued in `enqueue_block_editor_assets` the same as it is in `wp_enqueue_scripts`.

oxyc:
> editor.js should not be enqueued on the frontend.

No of course not, this PR only registers it.

If you prefer duplicating the `wp_enqueue_script` line rather than separating registrations from enqueues I'm fine changing the PR.

As for leaving it as it is, I'd vote no as you can't write es6 code without having to manually add `vendor.js` as a dependency. Something which isn't very straightforward to figure out since the "bug" doesn't issue any warnings/notices and comes from a babel runtime not mentioned anywhere in the sage code.

Log1x:
> No of course not, this PR only registers it.

Ahh I see now. Sorry, I had initially replied on my phone.

> As for leaving it as it is, I'd vote no 

I agree with you. We should add it in, even if commented out by default (open to opinions) – at least until we have vendor splitting between the app and editor.

I would like to stick to using the original method of `wp_enqueue_script`, though.

More or less, just keep the original but change to something along the lines of (untested):

```php
add_action('enqueue_block_editor_assets', function () {
    if ($manifest = asset('js/manifest.asset.php')->get()) {
        wp_enqueue_script('app/vendor.js', asset('js/vendor.js')->uri(), $manifest['dependencies'], $manifest['version']);
        wp_enqueue_script('app/editor.js', asset('js/editor.js')->uri(), $manifest['dependencies'], $manifest['version']);

        wp_add_inline_script('app/vendor.js', asset('js/manifest.js')->contents(), 'before');
    }

    wp_enqueue_style('app/editor.css', asset('css/editor.css')->uri(), false, null);
}, 100);
```

oxyc:
@Log1x I updated the PR but can't really see a good way to have it commented out since:

1. I think we should have it as a dependency of `sage/editor.js` otherwise I can potentially see some third party plugin breaking things. If you disagree I can of course remove it as a dependency.
2. As the inlined manifest needs come before `sage/vendor.js` we'd have to provide two version of that too.

Log1x:
I'm sorry! I completely forgot about this PR til today. I have added similar changes to #2413 so I'm closing this as it'll get merged it with that.

PR #2372, opened 2019-12-25
Keep hierarchical assets directory structure

oxyc:
### Issue summary

When organizing image and font assets in subdirectories their paths are flattened eg. `resources/assets/images/foo/bar.png` -> `dist/assets/images/bar.png`

With this PR it becomes `dist/assets/images/foo/bar.png`

### Solution

At first I thought this was a bug in the `laravel-mix-copy-watched` package ([invalid bug report](https://github.com/dsktschy/laravel-mix-copy-watched/issues/10)) but turns out it works as intended and keeping the hierarchy is actually an additional option.

I thought I'd open up a PR as others might encounter this as well. This is how it worked in sage 9, no?

Feel free to close this if you don't agree that this is a better default.

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -26,8 +26,8 @@ mix.js('resources/assets/scripts/app.js', 'scripts')
    .blocks('resources/assets/scripts/editor.js', 'scripts')
    .extract();
 
-mix.copyWatched('resources/assets/images', 'dist/images', {base: 'resources/assets/images'})
-   .copyWatched('resources/assets/fonts', 'dist/fonts', {base: 'resources/assets/fonts'});
+mix.copyWatched('resources/assets/images/**', 'dist/images')
+   .copyWatched('resources/assets/fonts/**', 'dist/fonts');
 
 mix.autoload({
   jquery: ['$', 'window.jQuery'],
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -26,8 +26,8 @@ mix.js('resources/assets/scripts/app.js', 'scripts')
    .blocks('resources/assets/scripts/editor.js', 'scripts')
    .extract();
 
-mix.copyWatched('resources/assets/images', 'dist/images')
-   .copyWatched('resources/assets/fonts', 'dist/fonts');
+mix.copyWatched('resources/assets/images', 'dist/images', {base: 'resources/assets/images'})
+   .copyWatched('resources/assets/fonts', 'dist/fonts', {base: 'resources/assets/fonts'});
 
 mix.autoload({
   jquery: ['$', 'window.jQuery'],


Log1x:
It is intended for the paths to be flattened and was the same behavior on Sage 9 (AFAIK).

oxyc:
Just tested to make sure, Sage 9 didn't flatten the path.

Log1x:
Gotcha.

In this case, it is intended to be flattened.

dylanjameswagner:
@Log1x What happens if there is more than one file with the same file name? Flattening would lead to a file name collision. What is the intended solution in this case?

Log1x:
The files and folders inside of the asset folders are not flattened. They just don't live in an asset directory inside of dist.

dylanjameswagner:
@Log1x Thank you for the clarification, I misunderstood the original intent of the PR.

kellymears:
For what it's worth I just SSHed into a live Sage 9 project (rebuilt this morning) and checked out the `dist` folder. Output of `tree`:

```
.
├── assets.json
├── images
│   ├── new-florida-majority-splash_36e8a018.jpg
│   ├── new-fm-logo_c9632e96.png
│   └── nfm-horizontal_8bf85334.png
├── scripts
│   ├── customizer_f67555a9.js
│   └── main_f67555a9.js
└── styles
    └── main_f67555a9.css
```

oxyc:
> For what it's worth I just SSHed into a live Sage 9 project (rebuilt this morning) and checked out the dist folder.

This doesnt match my test in the current master branch O.o

```
$ tree resources/assets/images/                                                                                                                                   5s 598ms
resources/assets/images/
└── test
    └── logo.png

1 directory, 1 file
$ tree dist/images/
dist/images/
└── test
    └── logo.png
```

> > @Log1x What happens if there is more than one file with the same file name? Flattening would lead to a file name collision. What is the intended solution in this case?
>
> The files and folders inside of the asset folders are not flattened. They just don't live in an asset directory inside of dist. It is more or less a redundant folder.

This is an issue in dist though because of the flattening.

```
 $ tree resources/assets/images/
resources/assets/images/
├── logo.png
└── test
    └── logo.png

1 directory, 2 files
$ tree dist/images/
dist/images/
└── logo.png
```

Log1x:
I see what you are saying now. You have me initially confused as it seemed like you were suggesting that compiled assets should live in an `assets` folder within `dist` which is not the case (nor was it on Sage 9).

You're right about copyWatched flattening the directories, though.

This can be fixed by passing `**` to the `from` directory like so:

```
mix.copyWatched('resources/assets/fonts/**', 'public/fonts')
   .copyWatched('resources/assets/images/**', 'public/images');
```

oxyc:
Oops, just re-read my issue, sorry about the confusion

oxyc:
The modified version you merged doesn't fix this issue.

```
$ tree resources/assets/images/                                                                                                                                       
resources/assets/images/
├── bar
│   ├── baz
│   │   └── foo.png
│   └── foo.png
└── foo.png

$ tree dist/images/
dist/images/
└── foo.png

```

PR #2409, opened 2020-02-26
Add `Requires PHP` and `Requires at least`; Bump WordPress requirement to v5.4

TangRufus:
See: https://make.wordpress.org/core/2020/02/26/miscellaneous-developer-focused-changes-in-wordpress-5-4/

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ See a working example at [roots-example-project.com](https://roots-example-proje
 
 Make sure all dependencies have been installed before moving on:
 
-* [WordPress](https://wordpress.org/) >= 5.2
-* [PHP](https://secure.php.net/manual/en/install.php) >= 7.1.3 (with [`php-mbstring`](https://secure.php.net/manual/en/book.mbstring.php) enabled)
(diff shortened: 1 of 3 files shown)


QWp6t:
We can remove this functionality from functions.php now.

https://github.com/roots/sage/blob/10.0.0-dev/functions.php#L21-L33

TangRufus:
If we remove https://github.com/roots/sage/blob/7a343ad8e6cdd861820879415e00dcbfe372f7f2/functions.php#L21-L33, then we need to bump WordPress requirement to v5.4.

Is that okay?

QWp6t:
We've never shied from bumping version requirements in Sage since it's a starter theme, so it's anticipated that devs are working with a new installation. I'm fine with bumping to WP 5.4, but we can wait to see what @retlehs thinks.

retlehs:
fine with me :+1:

TangRufus:
Note that the target date for release of WordPress 5.4 is 31st March 2020.
See: https://make.wordpress.org/core/5-4/

PR #2421, opened 2020-03-09
Get theme as disk

tylerwiegand:
The theme disk is too difficult to get to. Fix it!

I don't really intend to use the URL thing with it, it's just to have easy access to the Storage facade to write files when a folder isn't there. Maybe there's a better way to do this, but I've been using it and haven't found a downside; maybe someone has a better way!


config/filesystems.php
```diff
@@ -57,6 +57,13 @@
             'url' => site_url(),
             'visibility' => 'public',
         ],
+        
+        'sage' => [
+            'driver' => 'local',
+            'root' => get_theme_file_path(),
+            'url' => site_url(),
```
QWp6t, line 64: This should be `get_theme_file_uri()` to be consistent with `get_theme_file_path()`.

retlehs:
> @retlehs changed the base branch from 10.0.0-dev to master 

well, this destroyed this PR. do you mind re-submitting this against master now that sage 10 is in the main branch?

Log1x:
Added in #2520

PR #2435, opened 2020-04-05
Add html and body to PurgeCSS whitelist

darrenjacoby:
Purgecss removes styling for html and body which will have an impact of normalise/resets, add both to whitelist. https://github.com/FullHuman/purgecss-webpack-plugin/issues/69

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -20,7 +20,7 @@ mix.setPublicPath('./dist')
 mix.sass('resources/assets/styles/app.scss', 'styles')
    .sass('resources/assets/styles/editor.scss', 'styles')
    .purgeCss({
-     whitelist: require('purgecss-with-wordpress').whitelist,
+     whitelist: ['html', 'body', require('purgecss-with-wordpress').whitelist],
      whitelistPatterns: require('purgecss-with-wordpress').whitelistPatterns,
    });
 


MWDelaney:
This is still removing some block styles for me. I'm not sure how to usefully show that.

oxyc:
I think it's just that `index.php` isn't checked by purgecss. See list of [defaults paths](https://github.com/spatie/postcss-purgecss-laravel/blob/master/src/defaultConfig.js#L2:L12).

```js
   .purgeCss({
     extend: {
       content: [path.join(__dirname, 'index.php')],
     },
   })
```

Log1x:
I'm kind of torn on this whole situation right now. 

The reason the styles aren't being purged is because I extracted a majority of `app.blade.php` back to `index.php` due to https://github.com/roots/sage/issues/2116

Generally, I am unhappy at the idea of not using Blade for everything. Moving everything to `index.php` effectively kills the efficiency of Blade features such as [Stacks](https://laravel.com/docs/7.x/blade#stacks) (assuming you want them in your `<head>`, etc.)

While the approach used in Sage 9 [isn't the prettiest](https://github.com/kimhf/sage/blob/435e9f26dc8617100b94b5b45ec9239def5ac00e/app/filters.php#L50-L58) – it _is_ still an option to just do the same in Sage 10 (behind the scenes in Acorn).

The only thing I'm unsure of is if/is there is any noteworthy technical debt of the Sage 9 approach.

raffjones:
I just encountered this very issue. Lots of styles were missing when I ran the production build - it was not finding header styles etc. Easily fixed by extending the purgeCss as described by @oxyc but it might be good to extend out to include `^front-page.blade.php` for custom homepages, and to make sure standard navigation area classes are captured - I couldn't see any partials in the `whitelistPatterns` array.

Log1x:
> I just encountered this very issue. Lots of styles were missing when I ran the production build - it was not finding header styles etc. Easily fixed by extending the purgeCss as described by @oxyc but it might be good to extend out to include `^front-page.blade.php` for custom homepages, and to make sure standard navigation area classes are captured - I couldn't see any partials in the `whitelistPatterns` array.

All of `resources/` is already covered by https://github.com/spatie/postcss-purgecss-laravel/blob/master/src/defaultConfig.js

raffjones:
> All of `resources/` is already covered by https://github.com/spatie/postcss-purgecss-laravel/blob/master/src/defaultConfig.js – only `index.php` should have this issue due to it being in the theme root.

@Log1x Thanks for the pointer. I'm not 100% sure I understand what you mean - ie. if this is already supposed to be present in Sage 10. The Sage 10 `webpack.mix.js` doesn't include that, only `purgecss-with-wordpress`, which doesn't include any of these patterns. Not sure how to edit that file to include them.

Log1x:
The rules are bundled into `mix.purgeCss()` it's self

https://github.com/spatie/laravel-mix-purgecss

raffjones:
> The rules are bundled into `mix.purgeCss()` it's self so they are added no matter what. `purgecss-with-wordpress` is just appended to them.
> 
> https://github.com/spatie/laravel-mix-purgecss

OK - well, it isn't working for me, strangely (fresh install of Sage 10). No styles in any partial were found out of the box.

Log1x:
You might have to make a few whitelist rules depending on your application and where styles are rendered (e.g. plugin styles would need their own pattern whitelisted, etc.)

There is the option of whitelisting in the CSS directly https://purgecss.com/whitelisting.html#in-the-css-directly

Otherwise, as far as `front-page.blade.php` and other partials not working – I am slightly confused on why it wouldn't be working for you out of the box.

Screenshot

raffjones:
Thanks again for the pointers - I will dig into this tomorrow. Certain things just won’t work - for example a custom icon font. But I’m sure I’ll be able to figure it out thanks to your guide in the right direction.

raffjones:
@Log1x tl;dr There is no inherent issue with Sage 10 here, and the purgeCss rules are fine.

Thanks again for your pointers. I am using Roots Soil, which changes the name of all menu item classes to `menu-*`, and I'm using a custom icon font which take the form `<span class="icon icon-instagram">` for example. Adding this to `webpack.mix.js` resolved those problems and the page renders as expected:

````
.purgeCss({
    extend: {
      content: [path.join(__dirname, 'index.php')],
      whitelistPatterns: [/^menu(-.*)?$/, /^icon(-.*)?$/],
    },
   });
````

I do think that the unexpected nature of styles missing for the first time on `build:production`, followed by the learning curve of having to write patterns etc. for PurgeCSS will cause many users just to think the theme isn't working. Just an opinion based on this experience.

That said, it works incredibly well and I love it. My CSS file is 10k despite a full import of Bootstrap in the SCSS (I used to have to break out the `@import` rules to just bring in what I need but even then there was awful bloat).

Log1x:
Glad you got it working!

> I do think that the unexpected nature of styles missing for the first time on build:production, followed by the learning curve of having to write patterns etc. for PurgeCSS will cause many users just to think the theme isn't working. Just an opinion based on this experience.

I agree. Once the documentation is written, I'm hoping we will be able to make a few "gotchas" like that very clear from the beginning.

Log1x:
Sorry for the bump – as far as the menu styles go, check out https://github.com/log1x/navi – it will save your sanity & solve the need to whitelist the menu rules by having the markup in a view.

raffjones:
> Sorry for the bump

No need for apologise, always glad to get more info - I'll look at your repos for goodies.

PS: Do you never sleep? ;)

Log1x:
Added @oxyc's implementation in #2520

PR #2459, opened 2020-04-29
[9.x] Add support for efficiently inlining SVGs

strarsis:
This PR adjusts the webpack config to efficiently inline SVGs into CSS ([following this webpack documentation](https://webpack.js.org/loaders/url-loader/#svg)).
`data:image/svg+xml,` is used instead `base64` which reduces the overhead significantly.
It also updates two loaders (`css-loader` and `url-loader`) to add the required options and compatibility.
The `node-sass` package is also updated to make `npm` install on platforms like WSL working.


strarsis:
Hm, node v8 isn't compatible with the new releases of these two loader packages, hence the CI build fails.

PR #2504, opened 2020-07-03
use @roots/bud

kellymears:
This PR replaces laravel-mix with @roots/bud.


app/setup.php
```diff
@@ -32,14 +32,10 @@
  * @return void
  */
 add_action('enqueue_block_editor_assets', function () {
-    if ($manifest = asset('scripts/manifest.asset.php')->get()) {
-        wp_enqueue_script('sage/vendor.js', asset('scripts/vendor.js')->uri(), $manifest['dependencies'], null, true);
-        wp_enqueue_script('sage/editor.js', asset('scripts/editor.js')->uri(), ['sage/vendor.js'], null, true);
+    wp_enqueue_script('sage/vendor', asset('vendor.js')->uri(), ['edit-post', 'dom-ready', 'blocks'], null, true);
+    wp_enqueue_script('sage/editor/js', asset('scripts/editor.js')->uri(), ['sage/vendor'], null, true);
 
-        wp_add_inline_script('sage/vendor.js', asset('scripts/manifest.js')->contents(), 'before');
```
kellymears, line 39: need to add this back. was just trying to figure out the splitChunks issue.

Log1x:
Let's get something implemented in Acorn to handle the manifest stuff a bit more elegantly. 

cc: @QWp6t

bud.config.js
```diff
@@ -1,11 +1,8 @@
-/**
- * Bud: asset management framework.
- * @see https://roots.github.io/bud-support
- */
-const bud = require('@roots/budpack');
+// @ts-check
+const {bud} = require('@roots/bud');
```
kellymears, line 2: You'll need to use yarn link or some other solution to alias the package until it is published.

kellymears:
@retlehs found the cause of the `node_modules` bloat. 

It was a single import: `@babel/preset-modules`. Over 1gb 😲! I had taken it on when I was trying to see if I could dogfood bud.

https://github.com/roots/bud-support/blob/6f1952a1674bfa5f8320fa89efe8287435a7f67e/packages/bud/package.json#L41

Removed in `@2.0.0-next.16`. Modules back down to 350MiB. I saw a couple other redundant dependencies, but it's gonna be diminishing returns in comparison.

Screen Shot 2020-09-10 at 5 38 51 AM

sage.config.js
```diff
@@ -0,0 +1,72 @@
...
+    bud.src('styles/app.scss'),
+  ])
+  .bundle('editor', [
+    bud.src('scripts/editor.js'),
+    bud.src('styles/editor.scss'),
+  ])
+  .bundle('customizer', [
+    bud.src('scripts/customizer.js'),
+  ])
+  .copy('{images,fonts}/**/*')
```
kellymears, line 37: I like this change in a big way but I need to check that there are results for the glob before passing it to webpack. right now you'll get an error if there is no match. remove this line or add some files to those dirs if you're testing.

app/setup.php
```diff
@@ -14,16 +14,22 @@
...
 
-    wp_add_inline_script('sage/vendor.js', asset('scripts/manifest.js')->contents(), 'before');
+    /**
+     * Needs working out.
+     */
+    wp_enqueue_script('sage/vendor/app', get_home_url() . $manifest['vendor/app.js'], ['jquery'], null, true);
+    wp_add_inline_script('sage/vendor/app', asset('runtime/app.js')->contents(), 'before');
+    wp_enqueue_script('sage/vendor/app~customizer', get_home_url() . $manifest['vendor/app~customizer.js'], ['sage/vendor/app'], null, true);
+    wp_enqueue_script('sage/vendor/app~editor~customizer', get_home_url() . $manifest['vendor/app~editor~customizer.js'], ['sage/vendor/app~customizer'], null, true);
+    wp_enqueue_script('sage/app', get_home_url() . $manifest['app.js'], ['sage/vendor/app~customizer'], null, true);
```
kellymears, line 26: It is not my intent for these changes to be permanent.

Ultimately, I think the entrypoints manifest should be readable from acorn and this space can be way more removed from the wordpress api (maybe something like `enqueueEntrypoint('app.js')`?) . Currently the entrypoints plugin generates this output:

**entrypoints.json**

```json
{
  "app": {
    "runtime/app": "/app/themes/sage/dist/runtime/app.72976a14.js",
    "vendor/app~customizer": "/app/themes/sage/dist/vendor/app~customizer.72976a14.js",
    "vendor/app": "/app/themes/sage/dist/vendor/app.72976a14.js",
    "app": "/app/themes/sage/dist/app.72976a14.css"
  },
  "editor": {
    "runtime/editor": "/app/themes/sage/dist/runtime/editor.72976a14.js",
    "editor": "/app/themes/sage/dist/editor.72976a14.css"
  },
  "customizer": {
    "runtime/customizer": "/app/themes/sage/dist/runtime/customizer.72976a14.js",
    "vendor/app~customizer": "/app/themes/sage/dist/vendor/app~customizer.72976a14.js",
    "customizer": "/app/themes/sage/dist/customizer.72976a14.js"
  }
}
```

Totally willing to change that if @QWp6t needs it structured differently.

Because of the proxy built into the dev server a missing asset won't return a 404 in development. This is why there are no acorn errors when running a dev build even though the css isn't written to disk.

Log1x: I think the best solution would be to put the filetypes in an array so we can loop it and remain using WordPress' enqueue functions.

From Encore:

```
{
  "app": {
    "js": [
      "modules/dashboard/runtime.js",
      "modules/dashboard/vendor.js",
      "modules/dashboard/app.js"
    ],
    "css": [
      "modules/dashboard/vendor.css",
      "modules/dashboard/app.css"
    ]
  }
}
```

kellymears: @QWp6t mentioned also wanting arrayed assets. how do people feel about this:

```json
{
  "entrypoints": {
    "app": {
      "js": {
        "runtime/app": "/app/themes/sage/dist/runtime/app.72976a14.js",
        "vendor/app~customizer": "/app/themes/sage/dist/vendor/app~customizer.72976a14.js",
        "vendor/app": "/app/themes/sage/dist/vendor/app.72976a14.js",
        "app": "/app/themes/sage/dist/app.72976a14.js"
      },
      "css": [
        "/app/themes/sage/dist/app.72976a14.css"
      ]
    }
  },
  "assets": [
    "/app/themes/sage/dist/images/image.png",
    "/app/themes/sage/dist/fonts/font.otf",
    "/app/themes/sage/dist/whatever-assets-arent-js-or-css.svg"
  ]
}
```

And, is it helpful to have absolute paths in addition to public root relative ones?

```json
{
  "entrypoints": {
    "app": {
      "js": {
        "runtime/app": {
          public: "/app/themes/sage/dist/runtime/app.72976a14.js",
          path: "/absolute/path/to/wp/app/themes/sage/dist/runtime/app.72976a14.js",
        }
      }
    }
  }
}
```

I could see that being most useful with the runtime chunks, since this would undo any need for the PHP to derive a path from the public URL.

Log1x: I'd say no to absolute path since it's a publicly readable file and I think it's weird exposing stuff like that about my web application – but maybe I'm weird. – otherwise the entrypoints looks 🔥

QWp6t: For the assets, I was thinking of having them keyed by their relative path without cache-busting, similar to how we've always done it. And the CSS files should have a key so they can be named to ensure that if two entrypoints are loaded on a page and both share a CSS file, the CSS file will only be loaded once.

Also, I was hoping we could add the external dependencies for each entrypoint. I've already written something similar to work with encore so if you don't already have access to this, I can look into getting it into bud next week.

One more suggestion: the option to have the URL values be relative to the theme instead of root. This would allow developers to more easily distribute the theme separate from WordPress. That's what my reasoning behind it was in Sage 9.

In this example, I put the external dependencies in the entrypoints as well as a separate key on the bottom. We only need one, and it doesn't really matter which. So I'm cool with whatever. This is just to demonstrate what I'm talking about.

```json
{
  "entrypoints": {
    "app": {
      "js": {
        "runtime/app": "dist/runtime/app.72976a14.js",
        "vendor/app~customizer": "dist/vendor/app~customizer.72976a14.js",
        "vendor/app": "dist/vendor/app.72976a14.js",
        "app": "dist/app.72976a14.js"
      },
      "css": {
        "app": "dist/app.72976a14.css"
      },
      "dependencies": ["wp-blocks", "wp-dom-ready", "wp-element", "wp-hooks"]
    }
  },
  "assets": {
    "images/image.png": "dist/images/image.1a2b3c4d.png",
    "fonts/font.otf": "dist/fonts/font.1a2b3c4d.otf",
    "whatever-assets-arent-js-or-css.svg": "dist/whatever-assets-arent-js-or-css.1a2b3c4d.svg"
  },
  "dependencies": {
    "app": ["wp-blocks", "wp-dom-ready", "wp-element", "wp-hooks"]
  }
}
```

Let me know your thoughts. Like I mentioned in Slack, I'm totally fine if everyone isn't sold on the idea of having a unified manifest.json file.

.env.example
```diff
@@ -0,0 +1,5 @@
+APP_NAME='Sage'
+APP_ENV='development'
+APP_SSL=false
+APP_HOST='example.test'
+APP_PUBLIC_PATH='app/themes/sage/dist'
```
knowler, line 5: I am a bit hesitant with the naming of this file since it could give the impression that secrets go in here which could be devastating for since the theme is public.

QWp6t: Do you think it would be sufficient to simply add a warning at the top?

We could be obnoxiously loud about it:

```
# ⚠ ⚠ ⚠ DO NOT STORE SECRETS HERE!
```

knowler: Ya, I thought about that too. It could work.

kellymears: This is a little more complex and a lot more flexible since v2 launch.

Current behavior:

  - Bud only includes env values prefixed by `APP_` in bundled code
  - Bud will explicitly ignore env values including `SECRET` in their property name

Examples:

- `APP_NAME` -- included in bundle
- `APP_SECRET_KEY` -- not included in bundle
- `SOME_UNRELATED_KEY` -- not included in bundle

app/setup.php
kellymears: Done. Basically just implemented @Log1x's structure: https://github.com/roots/bud/pull/25

Output: 

```json
{
  "foo": {
    "js": [
      "/runtime/foo.js",
      "/vendor/lodash.js",
      "/foo.js"
    ],
    "css": ["/foo.css"],
    "dependencies": ["wp-element"]
  },
  "bar": {
    "js": [
      "/runtime/bar.js",
      "/vendor/lodash.js",
      "/bar.js"
    ],
    "css": [],
    "dependencies": ["wp-element"]
  }
}
```

storage/bud/records.json
```diff
@@ -0,0 +1,100 @@
+{
```
kellymears, line 1: Since I've gotten a few questions about this:

I did intend to add this file to version control. This file is used by `@roots/bud-caching` to determine if modules in a build have changed since the build was last run. If the modules are still valid (there have been no changes to the config file or the changes to the config file did not modify the utilized modules) then it will skip needlessly recompiling those modules altogether.

In practice, this means that once this file is generated build time is significantly truncated. Further, Bud goes out of its way to not use absolute paths for resolving modules, so these path should be equally valid for development and CI. By keeping this file in version control Bud can utilize it in the context of a remote build and save time there, as well. 

If we don't want to keep this file in the Sage repo because it's super ugly, I appreciate that, but we should encourage users to not add this file to their .gitignore if they use a CI service.

kellymears:
Okay, the last major issue standing in the way of this is an issue with HMR and gutenberg. pmmmwh/react-refresh-webpack-plugin doesn't play nice with the window vars enqueued by wp. I think the fix is to disable the webpack externals handling in development and have the user supply the packages as dev dependencies. 

At the moment, I'm just using `wp` window globals in the `editor.js` script. HMR works great, though.

I included some barebones but functional asset management: https://github.com/roots/sage/blob/2820ae421b225df095da5395d93e24b8a749a45f/app/assets.php. I hope it can be helpful when working out how to enqueue everything in Acorn.

Feel free to nix the docker-compose.yml file. But I think it's helpful.

kellymears:
HMR issue is fixed.

Also new:
- `sage` tty bin replacing package.json scripts
- updated readme

I think that's pretty much it. Other parts of Bud might have issues but this feels stable to me.

resources/assets/styles/components/button.scss
```diff
@@ -22,6 +22,22 @@
         map-get($theme-colors, "primary"),
         map-get($theme-colors, "light")
       );
+
```
kellymears, line 25: I didn't mean to include this scss stuff. 

It does fix the block preview on our example gutenberg button but I'll open that as  a separate thing.

index.php
```diff
@@ -8,7 +8,7 @@
   </head>
 
   <body <?php body_class(); ?>>
-    <?php wp_body_open(); ?>
+    <?php // wp_body_open(); ?>
```
kellymears, line 11: This is uninentional.

.eslintrc.js
```diff
@@ -35,8 +35,8 @@ module.exports = {
...
         imports: 'always-multiline',
         exports: 'always-multiline',
-        functions: 'ignore'
-      }
-    ]
-  }
+        functions: 'ignore',
+      },
+    ],
+  },
```
kellymears, line 41: I owe this file a couple small additions. Ref https://github.com/roots/bud/blob/8a06e499a3d209dfb54a50b4aabb4500a3a67455/packages/sage/src/sage/eslint.ts#L4-L48

kellymears: Done.

index.php
kellymears: Reverted.

resources/assets/styles/components/button.scss
kellymears: Reverted.

sage.config.js
```diff
@@ -1,17 +1,13 @@
...
-      'resources/views/**/*',
-      'resources/assets/scripts/**/*',
-    ],
-    css: ['resources/assets/styles/**/*'],
-  });
-});
+/**
+ * Sage theme
+ */
+const theme = Sage.bootstrap();
```
kellymears, line 7: `@roots/sage` exports a `bootstrap` config and a `tailwind` config. They're wrapped so they don't conflict with another.

@Log1x to check out the one I did based on your theme, invoke `Sage.tailwind()`. I believe the rest of the file is the same between the two configs.

Config preset source for both in this dir: https://github.com/roots/bud/tree/stable/packages/sage/src/sage

kellymears:
Updated to 3.0.28, which adds the `@roots/bud-esbuild` and `@roots/bud-emotion` extensions.

- `@roots/bud-esbuild` is utilized by `@roots/sage` for production builds. 
- `@roots/bud-typescript`, `@roots/bud-babel` and `@roots/bud-react` fill in for `@roots/bud-esbuild` in development.
- This is because `esbuild` does not provide HMR by design. Fair enough. We can look to `snowpack` and `vite` for inspiration on how to solve for this in a more savvy way -- maybe.
- `@roots/bud-emotion` is not included with `@roots/sage`.

kellymears:
Closing in favor of #2643, which is way fresher.

PR #2557, opened 2020-10-19
Check for autoload.php in root vendor

freezy-sk:
Check for autoload.php not only in local vendor folder, but also in vendor folder inside Bedrock page root.
Requires roots/acorn/pull/75

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -11,7 +11,10 @@
 |
 */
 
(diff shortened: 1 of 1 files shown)


Log1x:
Would this not create conflicts by the autoloader in Bedrock being loaded twice?

PR #2582, opened 2020-11-12
Correct "trailingComma" value for prettier

theMosaad:
For "trailingComma" in `.prettierrc`, the value "true" is not accepted. Valid values are: "es5", "none", "all" https://prettier.io/docs/en/options.html#trailing-commas


knowler (approved):
Nice catch. Thanks!

PR #2594, opened 2020-12-09
[9.x] Bump sage-installer to 1.6.4 (Tailwind 2 support)

knowler:
The Sage 9 installer supports Tailwind 2 for its preset now (roots/sage-installer#40). This just bumps that dep in the Composer lock file.

**Update:** I made this a draft since I believe I should be adding Tailwind 2’s peer deps in the installer

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1710,16 +1710,16 @@
         },
         {
             "name": "roots/sage-installer",
-            "version": "1.6.3",
+            "version": "1.6.4",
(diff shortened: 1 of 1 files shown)


knowler:
Sidenote: I wonder if we ignored `composer.lock` then we wouldn’t need to make new releases every time we updated `sage-installer`? I know there are other non-dev deps that might need it though.

PR #2620, opened 2021-01-26
Renamed partials to underscore + partialname and changes single quotes to double

kevinblackdesk:
Renamed stylesheets to the SCSS partial naming convention, added the `_` in front of the names for the partials.

Note that `app.scss` and `editor.scss` do not include the `_` because these are not partials since these 2 files will get compiled into their own css file.


Log1x:
Latest version of Sage 10 doesn't really have any partials so this should now be entirely in the developers court on how they want to organize their styles.

PR #2622, opened 2021-01-26
Add "skip to content" link

MWDelaney:


resources/views/layouts/app.blade.php
```diff
@@ -1,7 +1,12 @@
 <div class="max-w-3xl mx-auto">
+
+  <a class="sr-only focus:not-sr-only" href="#main">
+    {{ __('Skip to content') }}
+  </a>
+
   @include('partials.header')
 
-    <main class="py-8 prose main">
+    <main id="main" class="py-8 prose main">
```
knowler, line 9: I’m having trouble finding legit research to back this up ([maybe this?](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/skip-to-main-content/)), but I usually include `tabindex="-1"` here as well. iirc it has to do with the focus being set onto the element.

MWDelaney: Sure, I'll add that.

knowler: Here’s a respected voice on the matter: https://css-tricks.com/how-to-create-a-skip-to-content-link/#comment-1755109

Not sure we are targeting any of the older browsers that it matters for, but it’s not like it’s adding a ton of bloat.

oxyc: Without `tabindex="-1"` when you click the link with `<return>` rather than `<ctrl> + <opt> + <space>` using voiceover, it doesn't actually move the focus in chrome so reading still continues from the top of the page

With `tabindex="-1"` focus is moved and reading continues from the expected area.

Not sure how real world users  click links though... Personally I just got confused since I expected `<return>` to work.

Issue #2631, opened 2021-02-10
Add dotenv file into project


fzhantw:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary

Environment variables could be set in ".env" file just like Laravel does.

## Motivation

For browsersync proxy url, I don't think it's a good idea to write it in "webpack.config.js",
because it would be track in git.
Using .env variable might avoiding local environment variable showing to other.

## Additional context

//

mrobit:
@fzhantw you can add a `config-local.json` in the same directory as `config.json` to set a custom proxy URL. This will be merged in with the `config.json` values via the `config.js` file that's used in the build process.

fzhantw:
Hi @mrobit , thanks for replying!
I'm working on Sage 10.
Unfortunately, "config.json" seems to be removed, so I wonder whether there's any mechanism to configure by environment?

Log1x:
See response on #2632

Log1x:
Also, check out https://laravel-mix.com/extensions/env-file

PR #2632, opened 2021-02-10
add .env file & set browsersync proxy url in it

fzhantw:
issue #2631


Log1x:
For the moment I'd prefer for this to be a change made by the developer and not part of the actual Sage core – but when #2504 comes to light – Bud will be organized like this by default.

PR #2643, opened 2021-02-28
Replace Laravel Mix with Bud

kellymears:


QWp6t:
It looks like Tailwind isn't working correctly. The configured classes aren't being included during compile, but all of the `@apply` stuff is.

I have confirmed that tailwind.config.js is loading during compile.

<details>
<summary>screenshots</summary>

image

image

image

</details>

kellymears:
@QWp6t I think this is almost 100% good to go. I don't think the base path is being set properly for relative css imports, but proxy/hmr/postcss/react all seem to be on their best behavior. File hashing works well, minification works well. It feels fast.

I will need to also try dropping some actual production dependencies. Out-of-the-box there is no runtime/vendor bundle generated as there are no deps to bundle.

- [ ] Fix base path for relative urls in css
- [ ] Confirm runtime / vendor bundles are correctly generated
- [ ] Fix CI

kellymears:
@Jamiewarb I really appreciate the offer! I've had my head down working on getting everything working with webpack@5, and that is finally over the finish line now. There is a lot that could be done over at [@roots/bud](https://github.com/roots/bud), and the project is in a much better place to accept contributions now that there isn't a giant PR threatening to rain all over everyone's work.

The most immediate need is for test coverage. But, I'd never turn down an exciting feature.

I believe @QWp6t has some changes he wants to make to [@roots/entrypoints-webpack-plugin](https://github.com/roots/bud/tree/stable/packages/@roots/entrypoints-webpack-plugin), which is used by [@roots/bud-entrypoints](https://github.com/roots/bud/tree/stable/packages/@roots/bud-entrypoints) but outside of that there is no longer anything in scope that is blocking this from being merged.

Jamiewarb:
Congratulations on getting it all working with webpack@5!

strarsis:
@Jamiewarb: I thought Sage 10 already uses webpack 5 (Laravel Mix)?

resources/scripts/components/index.js
```diff
@@ -0,0 +1,12 @@
...
+ * document has loaded.
+ *
+ * @param  {CallableFunction} onReady
+ * @return {void}
+ */
+export const domReady = (onReady) => {
+  window.requestAnimationFrame(function check() {
+    document.body ? onReady() : window.requestAnimationFrame(check);
+  });
+};
```
knowler, line 12:
```suggestion
export const domReady = (onReady) => {
  window.requestAnimationFrame(document.body ? onReady : domReady);
}
```

We can check for the body immediately, right? Do we even need to do that in an animation frame the first time it’s called? I don’t know enough about `window.requestAnimationFrame`, so I could be wrong.

As an aside, I think we should change the name of this function to `bodyIsReady` (alternatively prefixed with `my`), since it kind of breaks the mold of DOM readiness functions (particularly by going outside of the definition of `DOMContentLoaded`).

bud.config.js
```diff
@@ -0,0 +1,39 @@
...
+     */
+    .use(require('@roots/sage'))
+
+    /**
+     * Application entrypoints
+     */
+    .entry({
+      app: '**/app.{js,css}',
+      editor: '**/editor.{js,css}',
+      customizer: '**/customizer.{js,css}',
```
knowler, line 22: I think we should be more specific here for the entries instead of using globs. It’s nice that they do double-duty for getting both the js and css assets, and they are more terse to write, but I think they are less clear for Sage users who might not understand all that is happening.

```suggestion
      app: ['resources/scripts/app.js', 'resources/styles/app.css'],
      editor: ['resources/scripts/editor.js', 'resources/styles/editor.css'],
      customizer: 'resources/scripts/customizer.js',
```

retlehs (approved):

PR #2721, opened 2021-05-25
Make jQuery 3.5.1 a peer dependency

knowler:
Since Sage is relying on WordPress’ jQuery, we can save the user from needing to download it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "stylelint-config-standard": "^20.0.0",
     "tailwindcss": "^2.0.3"
   },
-  "dependencies": {
+  "peerDependencies": {
     "jquery": "^3.5.1"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "stylelint-config-standard": "^20.0.0",
     "tailwindcss": "^2.0.3"
   },
-  "dependencies": {
-    "jquery": "^3.5.1"
+  "peerDependencies": {
+    "jquery": "3.5.1"
```
austinpray, line 43: I expected the version range to still hold true 

```suggestion
    "jquery": "^3.5.1"
```

knowler: Do they keep it updated like that?

Log1x (approved):

Issue #2775, opened 2021-07-30
Sage 9 not working on Windows


kingkero:
* [X] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [X] This request isn't a duplicate of an existing issue
* [X] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

The theme isn't working on Windows because of hardcoded `/` instead of using `DIRECTORY_SEPARATOR`.

## Steps to reproduce

1. Have a Windows machine
2. Try to activate a theme that uses Sage 9.
3. Errors.

**Expected behavior:** [What you expect to happen]

It works.

**Actual behavior:** [What actually happens]

It can't load autoload.php or setup files.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

## Additional information

I can provide a PR for this later today.

retlehs:
thanks, but we only support windows through WSL

PR #2776, opened 2021-07-30
[9.x] Remove hardcoded slash, replace by DIRECTORY_SEPARATOR

kingkero:
fixes https://github.com/roots/sage/issues/2775


retlehs:
thanks, but we only support windows through WSL

PR #2777, opened 2021-07-30
Fix development instructions in README.md

### LINKED ISSUE #2656: Sage 10: White screen with PHP errors
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

At the moment when a PHP error occurs, it produces a white screen. The error is written to application.log but it would be helpful to have the error display on the screen. Another problem with this is that small errors that would usually let a site continue working completely disable the site. It would be good to have the error logging work in a similar way to Sage 9 or previous versions of Sage 10. 

Development is slow when you're having the check the log a lot and turning WP_DEBUG to false seems to have no effect. 

One further thing I noticed is that if you have an undefined variable, wrapping it in an `@if` still returns an undefined error.

Thanks for all your work on Sage so far. 

## Steps to reproduce

1. Install a fresh copy of the latest sage 10 dev
2. Write any sort of php which will deliberately return false or an undefined variable

**Expected behavior:** Show the error on screen or inline with the site

**Actual behavior:** White screen is displayed

**Reproduces how often:** 100%

## Versions

MacBook Pro (Mid 2015) – Catalina 10.15.7
Sage 10 (Latest dev)
NPM 5.6.0
Node 12.14.0
PHP 7.3.25
Laravel valet

eduardoarandah:
Without `WP_ENV` defined, we get white screen on php errors.

This constant is used in config/app.php.

fixes #2656

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Edit `app/setup.php` to enable or disable theme features, setup navigation menus
 
 - Run `yarn` from the theme directory to install dependencies
 - Update `webpack.mix.js` with your local dev URL
+- Add to wp-config.php `define('WP_ENV', 'development');`
 
 ### Build commands
 


knowler:
I don’t think this needs to be in the Sage README since it is more so a concern of the WordPress install (see Bedrock or a regular WordPress install).

PR #2837, opened 2021-11-01
Move compiled views back to cache dir by default

retlehs:


QWp6t (approved):

slackday:
One gotcha is that this new folder isn't git ignored in Bedrock

PR #2841, opened 2021-11-14
Upgrade to stylelint v14.x

jeffbyrnes:
* Use stylelint-config-recommended-scss, per [v14 migration guide on specifying how to handle SCSS & other CSS-like languages](https://github.com/stylelint/stylelint/blob/HEAD/docs/migration-guide/to-14.md#syntax-option-and-automatic-inferral-of-syntax)

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,5 +1,5 @@
 {
-  "extends": "stylelint-config-standard",
+  "extends": "stylelint-config-recommended-scss",
   "rules": {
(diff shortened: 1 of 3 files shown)


retlehs:
thanks for the PR! fyi, sage is now using stylelint v14 as of #2643

jeffbyrnes:
Well darn! Guess I missed that, esp. since I’m still back on Sage v9.x for now.

Issue #2700, opened 2021-05-06
Rename the master branch to main


huubl:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary

Rename the master branch to main

## Motivation

Following the defaults of Github.

## Additional context

[https://github.com/github/renaming](https://github.com/github/renaming)

PR #2850, opened 2021-11-22
README updates for Sage 10

knowler:
This updates the README to reflect changes made for Sage 10.

- Switches references to Laravel Mix to Bud.
  - I am using the Netlify URL for the docs. Not sure if we want to change this.
- While jQuery still works in Sage since it works in WordPress, I removed the mention of it as a feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Sage is a productivity-driven WordPress starter theme with a modern development 
 
 - Harness the power of [Laravel](https://laravel.com) and its available packages thanks to [Acorn](https://github.com/roots/acorn).
 - Clean, efficient theme templating utilizing [Laravel Blade](https://laravel.com/docs/master/blade).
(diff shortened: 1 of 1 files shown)


knowler:
@retlehs I had added a few of these things to the 10 release checklist, but I figured I could update them myself. The only thing I haven’t do is remove or change the whole `composer create-project` thing. Is that how we still want to recommend installing Sage?

retlehs:
@knowler thank you 🙏 

netlify URL is good until https://github.com/roots/bud/issues/683 lands, removing the mention of jQuery is good  :+1: 

the mention of browsersync needs to be removed, maybe @kellymears / @QWp6t have thoughts on specific verbiage to use instead

keeping the `composer create-project` instructions sounds good to me for now. i think now would be a good time to land #2700 as well.

knowler:
21c1eb2 @retlehs I removed the Browsersync mention and changed the Bud description to just use its tagline from the website.

retlehs (approved):

PR #2876, opened 2021-12-04
Use type declarations and bump minimum PHP version to 7.4

knowler:
This PR uses type declarations that are supported by PHP 7.4+ where it and therefore increases the project’s supported PHP version to 7.4.

Ref https://github.com/roots/bedrock/pull/619:

> PHP 7.3 EOL is 2021-12-06 
> https://www.php.net/supported-versions.php

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,7 +56,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[ci skip]')"
     strategy:
       matrix:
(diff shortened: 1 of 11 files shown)


retlehs:
let's pass on the type declarations for now (unless laravel 9.x is doing this), but can we get a separate PR opened for dropping PHP 7.3 support?

knowler:
Closed in favour of #2925.

PR #2917, opened 2021-12-30
chore: remove helpers.php

Log1x:
I've removed this on every project I've made with Sage 10 up to this point as I can't really find a reason I'd use a messy global function over using a method inside of a Composer, a trait, or my own class.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,6 @@ themes/your-theme-name/   # → Root of your Sage based theme
 │   ├── Providers/        # → Service providers
 │   ├── View/             # → View models
 │   ├── filters.php       # → Theme filters
-│   ├── helpers.php       # → Helper functions
 │   └── setup.php         # → Theme setup
 ├── composer.json         # → Autoloading for `app/` files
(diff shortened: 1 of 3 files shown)


retlehs (approved):

QWp6t (approved):

retlehs:
can we get the README and docs updated as well?

SergiArias:
> I've removed this on every project I've made with Sage 10 up to this point as I can't really find a reason I'd use a messy global function over using a method inside of a Composer, a trait, or my own class.
> 
> Thoughts?

I usually create functions inside helpers (not more than 5 per project) and I am interested to know deeper what would be the best approach. Maybe add a section in the documentation after this removal?

I understand that the simpler one is to add a method inside app/View/Composers/App so it's available to all the views.
In the case of a class/trait, where would we instantiate the class in the current folder structure? In bootstrap/app.php? In config folder? 

I understand it would be the same concept of acf composer which creates classes inside Blocks, Fields and Options folders and instantiate the classes through acf composer.

PR #2933, opened 2022-01-05
app/setup.php doesn’t need to be executable

knowler:
I just noticed at some point during the big Sage 10 PR `app/setup.php` became executable. I have no clue if this matters, but I figure we should just restore it to not be executable.


retlehs (approved):
:eyes: :pray:

PR #2941, opened 2022-01-10
fix(views): Use dot notation on entry-meta view includes

Log1x:
I have no idea how this went unnoticed for so long lol


kalenjohnson (approved):
I guess 😡

PR #2943, opened 2022-01-10
Use a button element for search form

knowler:
This is trivial. The only real benefit is that it might be easier to add an icon.


knowler:
Might need to escape this in blade. I forget.

QWp6t (approved):

retlehs (approved):
:+1: nice one

PR #2956, opened 2022-01-21
Dynamically set the Bud public path

markjaquith:
In the upgrade to Bud 5.1.0, the Bud public path was hardcoded as `/app/themes/sage/public/`.

Instead of requiring people to manually set their path, we should just build it from the location of the Bud config file using Node's `__dirname`.

### DIFF
--- a/bud.config.js
+++ b/bud.config.js
@@ -42,4 +42,4 @@ module.exports = (app) =>
     /**
      * Public path of application assets
      */
-    .setPublicPath('/app/themes/sage/public/');
+    .setPublicPath(__dirname + '/public/');


knowler:
Unfortunately, using Node’s `__dirname` for the public path won’t work, since the public path is used in asset URLs. For example, in Sage when a stylesheet references a font, it uses an absolute URL to the font instead of a relative one (in the built stylesheet). If we were to use `__dirname`, it would include the full filesystem path in that URL. We should be able to alleviate this if we just use relative URLs between assets in the first place. We’ll need to discuss that a bit.

retlehs:
with the next bud release, setting a `publicPath` will no longer be required :eyes:

ref https://github.com/roots/bud/pull/1008 /cc @kellymears

retlehs:
no longer needed as of bud v5.2.0 / #2959

PR #3002, opened 2022-03-09
🔥 remove default palette + gradients from theme.json

mike-sheppard:
Assuming most people using Sage will be defining their own palette, might be worth removing the defaults WP throws in there?

<table>
<tr>
<th>before</th>
<th>after</th>
<tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -11,7 +11,9 @@
         }
       ],
       "custom": false,
-      "customGradient": false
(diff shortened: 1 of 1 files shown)


retlehs (approved):
:+1: agreed, thank you for the PR!

PR #3019, opened 2022-03-27
Fix the style.css version

davideprevosto:
I suppose there is a typo in the style.css version attribute.


retlehs:
this is on purpose. folks pulling from `main` need to be aware that they're getting a dev version of sage. it's not `10.1.4` anymore once that tag has been published.

PR #3025, opened 2022-04-02
🔧 Add index.php to tailwindcss purge list

gabrielgiordan:
Since `index.php` is a part of the views and can have `tailwindcss` classes, it makes sense to have it included into the purge content list to not miss any css classes there.


retlehs (approved):
:+1: thank you!

PR #3039, opened 2022-05-11
🐛 Fix i18n warnings

strarsis:
This PR fixes the i18n warnings by adding a missing translators comment and uses a singular placeholder.
For single comments, `One` is used instead of a number (in a `comments title` context), which fixes the 2nd i18n warning.

Edit: Fixed something and merged commits into one, hence the force-pushes...


retlehs (approved):
thank you

PR #3076, opened 2022-07-08
Include `theme.json`; `patterns/` in translation scan command

strarsis:
It turns out that when using `--include` with `wp i18n` the `theme.json` file and the `patterns` directory 
[need to be explicitly included, too](https://github.com/wp-cli/i18n-command/issues/322#issuecomment-1124063949), as those would be skipped otherwise.

Edit:
The latest `wp i18n` version (>=[`2.4.0`](https://github.com/wp-cli/i18n-command/releases/tag/v2.4.0)) is needed for some bugfixes of otherwise blocking bugs.
Until `wp` CLI is also newly released with the new `wp i18n` version bundled, 
the latest `wp i18n` version can be installed independently from the `wp` CLI version by using 
`wp package install wp-cli/i18n-command:dev-main` 
and then just using it as usual (`wp i18n`).
Edit (October 2023): Current stable `wp cli` `2.8.1` contains a recent enough `wp i18n`.


strarsis:
@retlehs: Latest stable WP CLI appears to work fine with the used commands now, so this can be merged.

To make this more robust, `bud` should re-generate the `theme.json` before the `translate:pot` script.
This can be achieved by just calling the `build` script, but wouldn't be just re-generating the `theme.json` be better than building the whole project? Then `bud` needs a dedicated command that just re-generates the `theme.json` (when `theme.json` generation is enabled in `bud`, otherwise it can be a no-op).

strarsis:
@kellymears: A dedicated command for `bud` that re-generates the `theme.json` (or no-op if no `theme.json` generation is enabled) would be great for this! That command would run before the translation scan, for the translation source strings `theme.json` being up-to-date.

strarsis:
@retlehs, @kellymears: `wp cli` `2.8.1` contains a recent enough `wp i18n`, so this PR can be merged now!

PR #3095, opened 2022-10-14
Replace the yarn pot command by the one recommended in the docs

bonakor:
The yarn command was not able to really create a complete pot file. Most strings from the Blade templates were missing. With this command, it seems to be working as it should be.

I found the command in the docs https://docs.roots.io/sage/10.x/localization/#generating-language-files

I removed the previous command to avoid any misunderstanding.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
   "scripts": {
     "dev": "bud dev",
     "build": "bud build",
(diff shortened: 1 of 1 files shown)


retlehs:
See #3081, which I think we can merge now that WP-CLI released a new version

cc @strarsis

bonakor:
Sorry, I didn't realize https://github.com/roots/sage/pull/3081 fixed it already. While wp-cli teams handles the remaining bug, should this PR be merged as it has the working solution proposed in the docs? Or are the docs enough?

strarsis:
@retlehs, @bonakor: Trying with the latest stable WP CLI, the command appears to work now without installing edge/dev versions of WP CLI or the i18n package, so this can be merged now!

PR #3104, opened 2022-11-23
Aria label for link in header

blift:
Aria label added to header link to prevent accessibility issue.

### DIFF
--- a/resources/views/sections/header.blade.php
+++ b/resources/views/sections/header.blade.php
@@ -1,5 +1,5 @@
 <header class="banner">
-  <a class="brand" href="{{ home_url('/') }}">
+  <a class="brand" href="{{ home_url('/') }}" aria-label="{!! __('Home page', 'sage') !!}">
     {!! $siteName !!}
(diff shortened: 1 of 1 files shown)


knowler:
Hey @blift, I appreciate the care for accessibility, but I think this change would be best left to the developer using Sage. While, I can totally see the value of adding “Home page” or “home” to the link’s accessible name, I think it does potentially obscure the actual text of the link from assistive technology and that might also be undesirable. That’s a decision that I think teams using Sage should make themselves. I’m going to close this, but I’d be interested in any research findings you might have in regards to this sort of pattern.

blift:
Hey @knowler thank you for your reply, I think you are right it should be left to the developers, to minimize hardcoded strings. 

Below in section header there is aria-label for navigation but it's provided by wp_get_nav_menu_name function. So it make sense.

knowler:
@blift Ya, in that case, the `aria-label` is necessary since there are multiple `<nav>` elements on the page.

PR #3128, opened 2023-03-22
bump: bud@6.12.0 & improvements

kellymears:
- add: simple comments to `jsconfig.json`
- improve: `bud.wpjson.set` replacing `bud.wpjson.settings`.
  - `bud.wpjson.settings` is not deprecated for bud@7.0.0
  - the defaults for `bud.wpjson.settings` include disabling `dropCap`, which wasn't in the config before since `bud.wpjson.settings` fully overrides the defaults. Unless we explicitly set it in the config it will be `false` (see the `theme.json` diff).
- `if (import.meta.webpackHot)` seems safer than optional chaining, since `import.meta.webpackHot` isn't a real method. For instance `import.meta?.webpackHot?` will cause errors, which would be unexpected given our sample code.

**todo**

- [x] gotta actually bump to 6.12.0 after release
- [x] resolve `dropCap` question one way or the other (see above)


retlehs (approved):

kellymears:
This PR contains various improvements and cleanups to the configuration and JavaScript code. Key changes include:

- Adding simple comments to jsconfig.json.
- Improving config: `bud.wpjson.set` replaces `bud.wpjson.settings`. Note that `bud.wpjson.settings` is not deprecated.
- `bud.wpjson.settings` now disables `dropCap` by default.
- `app.js` and `editor.js` now use `if (import.meta.webpackHot)` instead of optional chaining, since `import.meta.webpackHot` is not a real method and using optional chaining might cause errors.

PR #3155, opened 2023-10-23
Fix for multiple .po files

sandrowuermli:
With multiple `.po` files (`de_DE.po`, `fr_FR.po`) it will show the following error:
```
Error: Too many positional arguments: ./resources/lang/fr_FR.po
error Command failed with exit code 1.
```

Seems like `update-po` only accepts one argument and doesn't handle the `*.po` properly.


retlehs:
A bit annoying that they don't accept a glob for that subcommand — thank you for the PR 🙏

PR #3167, opened 2023-12-19
🎨 Move default theme markup to the `app` layout

### LINKED ISSUE #3166: Admin bar breaking error pages
### Version

10

### What did you expect to happen?

The error pages should display correctly with the admin bar displayed. The option is checked by default on a clean Wordpress setup.

### What actually happens?

The layout of the error pages is broken when the admin bar is shown.

Here is an error page on a clean Sage 10 install. I just echoed an undefined variable in the App composer.
Capture d’écran 2023-12-18 à 16 55 39

When switching to incognito mode the error page looks fine.
Capture d’écran 2023-12-18 à 16 55 56

I also tried with Laravel Ignition with the same result.  
I found that the entire HTML of the page is sent into the "New" item menu as wa can see here:

Capture d’écran 2023-12-18 à 17 00 39

### Steps to reproduce

- Create a `roots/bedrock` project
- Create a `roots/sage` project in the themes directory
- Add a simple line of code that throws an error (e.g. `echo $test;` without declaring `$test`)

### System info

- macOS 12.6
- Laravel Valet with PHP 8.2

### Log output

_No response_

### Please confirm this isn't a support request.

Yes

### LINKED ISSUE #3171: do_action('get_header') should occur before wp_head()
### Version

10.7.0

### What did you expect to happen?

The default WordPress theme behavior is to load header.php using a call to `get_header()`. Sage doesn't do this but does include `do_action('get_header')` so that any plugins relying on that action (maybe to enqueue styles, scripts, or meta tags) still fire.

I've embedded a Gravity Form in a theme file using their [function call](https://docs.gravityforms.com/adding-a-form-to-the-theme-file/). For this to work, you need to [manually enqueue the required styles and scripts](https://docs.gravityforms.com/gravity_form_enqueue_scripts/) for the page where the form is appearing.

Gravity Forms is expecting the typical WordPress ordering, where the `get_header` action runs before `wp_head()`. The result should be that their styles get enqueued in `wp_head()`, before the Sage app.css.

### What actually happens?

Sage does fire the `get_header` action in index.php, but after `wp_head()` and after the body_open. The result for this Gravity Forms use case is that the styles don't get enqueued until the footer, which messes up the stylesheet ordering.

While not difficult to work around for my use case, I'm wondering if the ordering of that action should change in index.php. If any other plugins are relying on the `get_header` action to enqueue styles, scripts, or other things that belong in `wp_head()`, the current order doesn't work because it is too late. The `get_footer` action does have the proper ordering.

### Steps to reproduce

Use the `get_header` action to enqueue a style that should appear in the `<head>`. The style will appear in `wp_footer()` instead.

### System info

_No response_

### Log output

_No response_

### Please confirm this isn't a support request.

No

Log1x:
After testing with Debug Bar, it appears that moving the default theme markup to the `app` layout does not have any negative side-effects like before seen with Sage 9 (e.g. #2116). Another pair of eyes to confirm would be cool.

Alongside being cleaner to maintain this stuff in Blade, this also fixes a lot of inconsistencies/rendering issues with Acorn's exception handler.

## Change log

- 🎨 Move default theme markup to the `app` layout
- 🐛 Fix uncontrollable HTML output (Fixes #3166)
- 🔧 Remove `index.php` from the Tailwind config
- 🩹 Move `get_header` hook above `wp_head()` (Fixes #3171)


QWp6t (approved):

retlehs (approved):
:+1: We'll ship this in the next Sage release once Acorn v4 is out

jeremylind:
I think `wp_head()` needs to be called in index.php outside the app context in order to properly enqueue scripts. In my testing, with a block with block.json defining a view script, the script is not enqueued when `wp_head()` is in the app layout.

PR #3181, opened 2024-02-01
Bump minimum PHP version to 8.1

dsturm:
In #3176 the minimum PHP version is bumped to 8.1 - updated this also in `style.css`


retlehs (approved):
Thank you!

PR #3189, opened 2024-05-14
🌰 Ship Acorn

MWDelaney:
Include Acorn with Sage by default.


retlehs:
@Log1x pointed out this needs the post autoload dump script added (either via `acorn:install` or [manually](https://roots.io/acorn/docs/installation/#add-the-autoload-dump-script))

MWDelaney:
Added!

Issue #3185, opened 2024-04-12
"editor.css" not included when Gutenberg (block) editor is in the new iframe mode


mjsarfatti:
### Version

10.8.2

### What did you expect to happen?

The new Gutenberg (block) editor is now inside an iframe. For this reason enqueueing editor assets using the `enqueue_block_editor_assets` hook no longer works reliably, and we are recommended to use `enqueue_block_assets` instead (see: https://github.com/WordPress/gutenberg/blob/trunk/docs/how-to-guides/enqueueing-assets-in-the-editor.md#editor-content-scripts-and-styles).

I expect styles in `editor.css` to be applied when the editor appears inside an iframe and I use the `enqueue_block_assets` hook.

### What actually happens?

Styles in editor.css are not applied in the following two cases:
1. When using the `enqueue_block_editor_assets` hook if the editor appears inside an iframe, or
2. When using the `enqueue_block_assets` hook both when the editor appears inside an iframe and when it appears in page (legacy).

### Steps to reproduce

1. Do a fresh install of Bedrock+Acorn+Sage and activate the Sage theme
2. Edit the file `web/app/themes/sage-theme/app/setup.php` like so:
```
- add_action('enqueue_block_editor_assets', function () {
+ add_action('enqueue_block_assets', function () { 
``` 
3. Add some CSS to `editor.css`, for example:
```css
.wp-block {
  background: pink;
}
``` 
4. Go to edit a page, and see that the background is not applied

### Please confirm this isn't a support request.

Yes

retlehs:
Both editor CSS and JS are working as expected for me with `enqueue_block_editor_assets` _and_ `enqueue_block_assets`

When using the current hook (`enqueue_block_editor_assets`) there is this warning:

```
editor/0-css was added to the iframe incorrectly. Please use block.json or enqueue_block_assets to add styles to the iframe.
```

I'll get a PR up with that change

AxelDeneu:
Any news on this?

On my side, `enqueue_block_editor_assets` works only when not previewing in responsive mode (with iframe).
As soon as an iframe is generated, the style doesn't apply.

Changing `enqueue_block_editor_assets` by `enqueue_block_assets` doesn't solve the issue as it removes every editor style.

Adding both `enqueue_block_editor_assets` and `enqueue_block_assets` with the same `bundle('editor')->enqueue();` doesn't solve this issue neither.

May this be an issue regarding the css encapsulation in the `editor.js` file?

[...]

PR #3196, opened 2024-09-19
🔧 Use `enqueue_block_assets` for editor assets

retlehs:
This PR fixes the following warning:

```
editor/0-css was added to the iframe incorrectly. Please use block.json or enqueue_block_assets to add styles to the iframe.
```

Ref #3185

### DIFF
--- a/app/setup.php
+++ b/app/setup.php
@@ -22,7 +22,7 @@ add_action('wp_enqueue_scripts', function () {
  *
  * @return void
  */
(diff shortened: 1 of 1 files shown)


imre08 (commented):
Before the editor assets were not loaded. After the change they are loaded. You will know when Tailwind CSS loads, then it resets Gutenberg editor styles.

Note that I did not see a warning before nor after the change.

PR #3204, opened 2025-01-15
Import static assets

tombroucke:
Similar to .assets(["images"]) in bud, we can add static assets in vite.
See [https://laravel.com/docs/11.x/vite#blade-processing-static-assets](https://laravel.com/docs/11.x/vite#blade-processing-static-assets)


retlehs:
Thank you 🔥

PR #3214, opened 2025-01-28
add built-in asset aliases

csorrentino:
add the same preset aliases as the @roots/sage bud extension


csorrentino:
closed as this doesn't need to be in the tailwind-v4 branch

PR #3215, opened 2025-01-29
add default aliases for assets

csorrentino:
Adds the same default aliases as the @roots/sage bud extension. Allows hashed asset paths to be referenced in css & js.
For example: ` background-color:url('@images/example.jpg');`

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -34,10 +34,10 @@ export default defineConfig({
   ],
   resolve: {
     alias: {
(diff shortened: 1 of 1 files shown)


retlehs (approved):
Thank you for the PR! Heads up - I pushed an update to remove the beginning `/` since they weren't needed in my testing, but please let me know if this breaks anything

I'll be getting some proper e2e tests setup soon, too

PR #3216, opened 2025-02-02
Fix theme.json generation to handle :host selector in app.css theme layer

stuart-james:
Previously, `wordpressThemeJson` checked that `themeContent` starts with  `:root{`, but  Tailwind wraps theme layer  in  `:root, :host`, causing an early return.  Remove '{'  to support `:host` selector and add `trim()` for non-minified app.css.


Log1x:
ty!

PR #3218, opened 2025-02-03
Fix default aliases paths for assets

lis-marcin:
Setting absolute paths in aliases to prevent warnings in Vite.js and fix the background-image reference to an image using eg. @images/bg.svg

before
after


retlehs (approved):
Thank you!

PR #3219, opened 2025-02-04
Fix Vite dependency extract

palicko:
Hey,
first of all, great job implementing Vite to the Sage, I love it!

Dependency extraction regex is working fine when imports are defined in single line, but not working when imports are defined using new lines, e.g.
```
import {
  InnerBlocks,
  InspectorControls,
  RichText,
  useBlockProps,
  useInnerBlocksProps,
} from '@wordpress/block-editor';
```
This PR should fix regex and add check if extracted import is not empty (last comma in import block was causing trouble).

Also I've added small fix for Typescript compatibility.

### DIFF
--- a/resources/js/build/wordpress.js
+++ b/resources/js/build/wordpress.js
@@ -88,7 +88,7 @@ export function wordpressPlugin() {
       }
     },
     transform(code, id) {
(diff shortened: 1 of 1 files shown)


retlehs (approved):
Thank you!

PR #3222, opened 2025-02-05
Compact fix for color generation in theme.json

codewithfeeling:
While testing I found that the `themeContent` CSS sometimes does not contain a closing curly bracket, meaning the index of the last bracket was -1. This cut off the final `--color-example: #abcdef` definition by one character, leading to the generated theme.json color for that one to be:

````
{
          "name": "example",
          "slug": "example",
          "color": "#abcde"
 }
````

By checking for the presence of the final bracket, we can either slice to there, or to the end of the content if it is not present.

````
const endIndex = themeContent.lastIndexOf('}');
const rootContent = themeContent.slice(themeContent.indexOf('{') + 1, endIndex === -1 ? undefined : endIndex);
````

### DIFF
--- a/resources/js/build/wordpress.js
+++ b/resources/js/build/wordpress.js
@@ -211,8 +211,8 @@ export function wordpressThemeJson({
       }
 
       const startIndex = themeContent.indexOf('{') + 1;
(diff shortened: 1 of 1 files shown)


retlehs (approved):
Thank you!

PR #3239, opened 2025-03-03
Add types

adevade:
This pull request includes various changes to improve type hinting and return type declarations. The changes ensure that functions and methods have explicit return types, improving code readability and maintainability.

### DIFF
--- a/app/Providers/ThemeServiceProvider.php
+++ b/app/Providers/ThemeServiceProvider.php
@@ -8,20 +8,16 @@ class ThemeServiceProvider extends SageServiceProvider
 {
     /**
      * Register any application services.
-     *
-     * @return void
      */
-    public function register()
+    public function register(): void
     {
         parent::register();
     }
 
     /**
      * Bootstrap any application services.
-     *
-     * @return void
      */
-    public function boot()
+    public function boot(): void
     {
         parent::boot();
     }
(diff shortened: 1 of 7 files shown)


Log1x:
I'm not against doing this for hooks but the changes to Composers and the default provider wouldn't work as they don't match the parent classes.

adevade:
@Log1x Well, you can always let the user land code be more specific than the parent classes, right?

Laravel does this for example:

[laravel/framework](https://github.com/laravel/framework/blob/56458795c1e68aa88bee5b415f5e40b67d482ed3/src/Illuminate/Support/ServiceProvider.php#L86-L94)
```php
abstract class ServiceProvider
{
    // ...

    /**
     * Register any application services.
     *
     * @return void
     */
    public function register()
    {
        //
    }

    // ...
}
```

[laravel/laravel](https://github.com/laravel/laravel/blob/540728ed96d84896c6d8c3ad263d64af2db831af/app/Providers/AppServiceProvider.php#L9-L15)
```php
class AppServiceProvider extends ServiceProvider
{
    /**
     * Register any application services.
     */
    public function register(): void
    {
        //
    }

    // ...
}
```

Log1x:
Fair point, but you still didn't test this at all.

```
Type of App\View\Composers\App::$views must not be defined (as in class Roots\Acorn\View\Composer)
```

PR #3245, opened 2025-03-13
🩹 Enable eager loading for images and fonts

retlehs:
Ensure that images and fonts are immediately imported rather than being lazily loaded to ensure that assets are available as URLs when needed

Ref https://discourse.roots.io/t/access-vite-import-meta-glob-image-url-in-js/29448

Will also be updating the https://roots.io/sage/docs/compiling-assets/ docs with an example

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -1,4 +1,1 @@
-import.meta.glob([
-  '../images/**',
-  '../fonts/**',
-]);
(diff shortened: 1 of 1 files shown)


Log1x:
We're going to stick to https://vite.dev/guide/features.html#static-assets for this use-case.

PR #3248, opened 2025-03-21
Always import the editor.css

johanmolen:
Always import the `editor.css` instead of inlining it when Vite is not running hot.

For example `@font-face` calls require an absolute URL to the font when inlined, due to being called from within an iFrame.
I guess the same issue appears for images referenced within the CSS file.

Instead of putting absolute URL's into the CSS file we can just import the CSS file, that way it uses the hostname of the CSS file.


Log1x (approved):

Log1x:
thanks!

PR #3270, opened 2025-08-07
fix tailwind source

kukac7:


retlehs:
@kukac7 What is this fixing?

kukac7:
@retlehs it has happened to me several times that certain class were not generated, even though they were in the code.

what's more, this is also the case in the [tailwind doc](https://tailwindcss.com/docs/installation/framework-guides/laravel/vite).

resources/css/app.css
```diff
@@ -1,3 +1,4 @@
 @import "tailwindcss" theme(static);
-@source "../views/";
-@source "../../app/";
+@source '../../storage/framework/views/*.php';
```
QWp6t, line 2: This makes sense but it means things will break if a user [changes where their compiled view files are located](https://roots.io/acorn/docs/directory-structure/). If possible, we should try to generate this path instead of hard coding it.

retlehs:
Pushed a small fix to your branch:

- Removed the cache source path (`../../../../cache/acorn/framework/views/*.php`) — those files may not exist in CI or clean builds
- Restored `app/` PHP scanning (`../../app/**/*.php`) — without this, Tailwind classes in Composers/Components won't be detected
- Kept your explicit Blade/JS globs and switched quotes to double for consistency

Also updated the PR title and description to better reflect the final change.

retlehs (approved):
Thank you!

PR #3274, opened 2025-10-06
🔧 Set APP_URL for Laravel Vite plugin

retlehs:
This prevents the "APP_URL: undefined" message when running the Vite dev server

* [ ] Update docs

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -3,6 +3,9 @@ import tailwindcss from '@tailwindcss/vite';
 import laravel from 'laravel-vite-plugin'
 import { wordpressPlugin, wordpressThemeJson } from '@roots/vite-plugin';
 
+// Set APP_URL for Laravel Vite plugin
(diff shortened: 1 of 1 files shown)


vite.config.js
```diff
@@ -3,6 +3,9 @@ import tailwindcss from '@tailwindcss/vite';
 import laravel from 'laravel-vite-plugin'
 import { wordpressPlugin, wordpressThemeJson } from '@roots/vite-plugin';
 
+// Set APP_URL for Laravel Vite plugin
+process.env.APP_URL = 'http://example.test';
```
QWp6t: I would check whether the value exists before setting it here.
```suggestion
// Set APP_URL if it doesn't exist for Laravel Vite plugin
if (! process.env.APP_URL) {
  process.env.APP_URL = 'http://example.test';
}
```

PR #3288, opened 2026-04-24
Support quotation marks in query field

tjernblom:
When using quotation marks in the query field, which is supported by Wordpress, the query is transformed to
`&quot;multiple words&quot;`
on the next page load. Easily fixed this way.


retlehs (approved):
Thank you!

PR #3291, opened 2026-06-12
fix: wrap get_permalink() with esc_url() in excerpt_more filter

thisismyurl:
The `excerpt_more` filter in `app/filters.php` places the result of `get_permalink()` directly into an `href` attribute without passing it through `esc_url()` first.

```php
// Before
return sprintf(' &hellip; <a href="%s">%s</a>', get_permalink(), __('Continued', 'sage'));

// After
return sprintf(' &hellip; <a href="%s">%s</a>', esc_url( get_permalink() ), __('Continued', 'sage'));
```

`get_permalink()` does not sanitize or escape its output. While `get_permalink()` constructs from trusted data in most cases, WordPress coding standards require all URLs placed into `href` attributes to pass through `esc_url()` — both for correctness and because a filter on `post_link` (or a malformed custom permalink structure) could introduce unescaped characters that `esc_url()` would catch and neutralize.

This is a one-line, safe change with no functional impact on normal output.

---

> **AI disclosure:** I use [Claude Code](https://claude.ai/code) as part of my development workflow. It helped me identify and verify this issue. I reviewed the code and the fix before opening this PR.


thisismyurl:
Good catch — pushed a follow-up that removes the extra spaces to match the file's existing style:

```php
esc_url(get_permalink())
```

